PR #16, opened 2015-05-25
WINDUP-611  Add XSD to all rules, improve the namespace

OndraZizka:
Hi,

the namespaces, filenames and URLs need to be informative and "scale" in the future.

1) It needs to say windup, so people know what namespace they are dealing with.

2) Same goes for the filename - when people download it. Therefore, rule-schema-1.0.xsd is not good IMO.

3) I suggest to align the XSD version with the Windup core version, since the XSD describes what the core accepts.
    "1.0" is not fortunate, as people will confuse it with legacy Windup.
    The version string should be the same as Windup version, so we can automate things. Different formats are source of human errors.

4) It doesn't need to say "schema" - XSD is saying that.

5) It's not rule - it's a ruleset. Could be misleading.

I suggest the URL for the current XSD to be:
    http://windup.jboss.org/schema/windup-ruleset-2.3.0.Final.xsd

And the namespace:
    Either the same,
    http://windup.jboss.org/schema/windup-ruleset-2.3.0.Final.xsd
    or,
    http://windup.jboss.org/ns/ruleset/2.3.0.Final , and that would be handled by the site (e.g., redirect to docs).

Sounds good?
Ondra


mareknovotny:
Do we need to have there full version string in schema and namespace? I would go with shorter version representation like http://windup.jboss.org/schema/windup-ruleset-2_3.xsd

My reasoning is micro releases should not change the schema and the rest is redundant there.

OndraZizka:
The version string should be the same as Windup version, so we can automate things. Different formats are source of human errors.

lincolnthree:
Using the version in the namespace is a bad idea, that doesn't make sense. Using the version name in the schema filename is a better solution. I agree with Jesse.

PR #17, opened 2015-05-26
WINDUP-612  Validate all XML rules in windup-rulesets during build.

OndraZizka:
Depends on https://github.com/windup/windup/pull/570

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -167,5 +167,50 @@
                 <directory>rules</directory>
             </resource>
         </resources>
+
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>Copy current ruleset schema from Windup core.</id>
+                        <goals>
+                            <goal>unpack</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>org.jboss.windup.config</groupId>
+                                    <artifactId>windup-config-xml</artifactId>
+                                    <version>${project.version}</version>
+                                    <type>jar</type>
+                                    <classifier>forge-addon</classifier>
+                                    <includes>**/*.xsd</includes>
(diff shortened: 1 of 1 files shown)


lincolnthree:
I think this is the wrong solution. I'd rather point the build to a hosted schema than pull it out of an artifact like this. Let's think about other solutions and discuss before implementing this.

OndraZizka:
But on meeting, you said that the authoritative place will be github repository, probably config-xml. This is the implementation of it.

mareknovotny:
We don't bundle schema into a jar/artifact for now, right? It would make sense to bundle it and validate to it from already downloaded Maven artifact

OndraZizka:
New PR to bundle the XSD in windup core:
https://github.com/windup/windup/pull/582

pom.xml
```diff
@@ -167,5 +167,50 @@
...
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>xml-maven-plugin</artifactId>
+                <version>1.0</version>
+                <executions> <execution> <goals> <goal>validate</goal> </goals> </execution> </executions>
```
lincolnthree: Formatting. Newlines needed here.

pom.xml
```diff
@@ -167,5 +167,50 @@
...
+                <configuration>
+                    <validationSets>
+                        <validationSet>
+                            <includes>
+                                <include>**/*.windup.xml</include>
+                            </includes>
+                            <dir>rules</dir>
+                            <systemId>${project.build.directory}/xsd/windup-ruleset-2.3.0.Final.xsd</systemId>
```
lincolnthree: This should reference the version-less schema.

pom.xml
OndraZizka: Changed. I will never understand why they are needed, though.

pom.xml
OndraZizka: Fixed.

OndraZizka:
Seems like I need to re-create the PR, github doesn't let me reopen.

PR #18, opened 2015-05-26
Add namespace and schema information to the example ruleset

sgilda:


lincolnthree:
Schema name should be windup-jboss.xsd

PR #19, opened 2015-05-27
Rules check ondra windup 610    WINDUP-610  Rules check - Ondra

OndraZizka:


lincolnthree:
Schema name needs to be fixed. Also, you need to include test cases for each rule. Additionally, please use complete, grammatically complete sentences in hints.

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
```diff
@@ -0,0 +1,1219 @@
...
+                </javaclass>
+            </when>
+            <perform>
+                <classification classification="WebSphere Asynchronous Work" effort="2">
+                    <link rel="sourceJavadoc" description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/api/content/nl/en-us/SS7JFU_7.0.0/com.ibm.websphere.javadoc.doc/web/apidocs/com/ibm/websphere/asynchbeans/Work.html"/>
+                </classification>
+                <hint message="`Work` is implemented by applications when they want to run code blocks asynchronously."/>
+                <!-- http://stackoverflow.com/questions/9026516/replacing-webspheres-workmanager-in-jboss -->
+                <hint>
+                    Depending on how what asyncbeans WorkManager is used for, it can be migrated to one of:
```
mbriskar: do we support hint as text node?

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
```diff
@@ -0,0 +1,1219 @@
...
+                     <link description="Servlet 3.0 Asynchronous Servlets" href="https://blogs.oracle.com/enterprisetechtips/entry/asynchronous_support_in_servlet_3"/>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="WebSphere>eap6:-Cbxnb">
+            <when>
+                <javaclass references="com.ibm.websphere.asynchbeans.WorkManager" as="default"/>
+            </when>
+            <perform>
+                <hint message="Migrate to JBoss JCA Work Manager." effort="1">
```
mbriskar: Is it enough to give a link to JCA Work Manager? Shouldn't we link to articles that tell you how to migrate?

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
```diff
@@ -0,0 +1,1219 @@
...
+        <rule id="WebLogic-weblogic.net.http.HttpsURLConnection.loadLocalIdentity()">
+            <when>
+                <javaclass references="weblogic.net.http.HttpsURLConnection.loadLocalIdentity()" as="default">
+                </javaclass>
+            </when>
+            <perform>
+                <hint message="Replace with `javax.net.ssl.HttpsURLConnection.getLocalPrincipal()`" effort="2"/>
+            </perform>
+        </rule>
+        <rule id="WebLogic-weblogic.net.http.HttpsURLConnection">
```
mbriskar: these IDs will need to be adjusted to yesterday's meeting

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
```diff
@@ -0,0 +1,1219 @@
...
+            <when>
+                <javaclass references="weblogic.net.http.HttpsURLConnection" as="default">
+                </javaclass>
+            </when>
+            <perform>
+                <hint message="Replace with `javax.net.ssl.HttpsURLConnection`" effort="2"/>
```
mbriskar: every hint will need to have severity

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
```diff
@@ -0,0 +1,1219 @@
...
+        </rule>
+        <rule id="WebSphere>eap6:MQeQC-10vvyf">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeQueueConnection" as="default"/>
+            </when>
+            <perform>
+                <classification description="Legacy IBM MQe API. A more general, and recommended, approach is to use the QueueConnection class.">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/jms/MQeQueueConnection.html"></link>
+                </classification>
+                <hint message="Replace with Java EE 6 JMS: `javax.jms.QueueConnection`." effort="1">
```
mbriskar: For example for such hints, I was putting JMS tag inside. Not sure it is correct

lincolnthree:
This PR still needs tests.

OndraZizka:
Sure, this is in progress, I will address all those notes above.

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
jsight: No, this needs a &lt;message&gt; element surrounding it.

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
jsight: I do think an article in this area would be helpful. Unless the WebSphere WorkManager is very similar to the standard one?

Also, I think the word "JBoss" is unnecessary in this title, as this is not a JBoss-specific API.

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
```diff
@@ -0,0 +1,1219 @@
...
+
+        <!-- Startup Service -->
+        <rule id="WebSphere>eap:StartUp-8jyqn">
+            <when>
+                <javaclass references="com.ibm.websphere.startupservice.{*}StartUp{*}" as="default">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification classification="WebSphere Startup Service" effort="4">
```
jsight: This should be 'title' now.

rules-reviewed/jboss-eap/websphere/WebSphere.windup.xml
jsight: I think that is a good idea, although it might also be a good idea to split all of the related rules into their own ruleset (something like websphere-jms.windup.xml).

OndraZizka: +1

OndraZizka:
Tests added.

pom.xml
```diff
@@ -163,8 +163,15 @@
     <build>
         <resources>
             <resource>
-                <filtering>false</filtering>
                 <directory>rules</directory>
+                <filtering>false</filtering>
+            </resource>
+            <resource>
+                <directory>rules-reviewed/jboss-eap/websphere</directory>
```
jsight: I don't think this belongs in the commit? Alternatively, maybe it could include all of rules-reviewed instead?

rules-reviewed/jboss-eap/websphere/WebSphere-other.windup.xml
```diff
@@ -0,0 +1,72 @@
...
+        <sourceTechnology id="websphere" versionRange="(10,12]"/>
+        <targetTechnology id="eap" versionRange="[6,)"/>
+        <targetTechnology id="java-ee" versionRange="[6,)"/>
+        <tag>reviewed-2015-06-WIP</tag>
+        <tag>ibm</tag>
+    </metadata>
+
+
+    <rules>
+        <rule id="WebSphereRules_6bcvk">
```
jsight: This should use the lowercase and '-' conventions.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                "com.ibm.mq.jms.context.WMQInitialContextFactory");
+        -->
+        <rule id="websphere-mq-r322f">
+            <when>
+                <!-- WINDUP-615
+                <javaclass references="com.ibm.mq.jms.context.WMQInitialContextFactory" >
+                    <location>STRING</location>
+                </javaclass>
+                -->
+                <filecontent filename="{*}.java" pattern="{*}&quot;com.ibm.mq.jms.context.WMQInitialContextFactory&quot;{*}"/>
```
jsight: Why only Java files?

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                    <link description="Migration Guide - HornetQ" href="https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6.4/html-single/Migration_Guide/index.html#Migrate_Your_Application_to_Use_HornetQ_as_the_JMS_Provider"/>
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+
+        <!-- .scp config files - WINDUPRULE-1 -->
+        <rule id="websphere-mq-fw4fw">
+            <when>
+                <filecontent filename="{*}.scp" pattern="{*}"/>
```
jsight: Good :)

rules-reviewed/jboss-eap/websphere/tests/WebSphereMQ.windup.test.xml
```diff
@@ -0,0 +1,322 @@
+<?xml version="1.0"?>
+<ruletest xmlns="http://windup.jboss.org/schema/jboss-ruleset">
+    <testDataPath>samples/WebSphereMQ</testDataPath>
+    <rulePath>../WebSphereMQ.windup.xml</rulePath>
+    <ruleset>
+        <metadata></metadata>
+        <rules>
+            <rule>
```
jsight: I think that tests should have ids, or at least some reference to the rule being tested.

rules-reviewed/jboss-eap/websphere/tests/samples/other/SimplePubSub.java
```diff
@@ -0,0 +1,83 @@
...
+ * Topics are dynamically created on the queue manager and need not be pre-defined.
+ *
+ * (The Broker must be enabled on the queue manager and the JMS Publish/Subscribe
+ * queues must be defined manually.)
+ *
+ * Note: These samples are for WMQ Base Broker and incomplete for WebSphere Message Broker)
+ *
+ * Does not make use of JNDI for ConnectionFactory and/or Destination definitions.
+ *
+ * @author saket
```
jsight: This appears to be a literal copy and paste from sample code on the web. Where is the license information?

src/test/java/org/jboss/windup/rules/tests/WindupRulesTest.java
```diff
@@ -288,6 +288,7 @@ private void runWindup(GraphContext context, File baseRuleDirectory, final List<
         windupConfiguration.addDefaultUserRulesDirectory(baseRuleDirectory.toPath());
 
         final String baseRulesPathNormalized = baseRuleDirectory.toPath().normalize().toAbsolutePath().toString();
+        System.out.println("baseRulesPathNormalized AAA " + baseRulesPathNormalized);
```
jsight: I am fine with this being logged, but could you format it a little better? :) I'm sure this was just debugging code that was left in, but I think it could be useful anyway.

jsight:
This rename shouldn't be here: rules-reviewed/jboss-eap/weblogic/weblogic.windup.xml → ...jboss-eap/weblogic/weblogic.windup-doesntwork.xml

pom.xml
OndraZizka: You're right. Whole rules-reviewed/ should be included.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Ok, filename= removed.

rules-reviewed/jboss-eap/websphere/tests/WebSphereMQ.windup.test.xml
OndraZizka: The reference is the hint being tested - quite easy to find, IMO.

rules-reviewed/jboss-eap/websphere/tests/samples/other/SimplePubSub.java
OndraZizka: It is a downloaded example app. Here is the license: http://www.ibm.com/developerworks/apps/download/index.jsp?contentid=48572&filename=StandaloneWMQCode.zip&method=http
I think it's ok. But I am no expert on licenses.

src/test/java/org/jboss/windup/rules/tests/WindupRulesTest.java
OndraZizka: Oh, I wanted to remove that :) Sure

OndraZizka:
Originally that file didn't really work, causing tests to fail. Now it's ok.

rules-reviewed/jboss-eap/websphere/tests/samples/other/SimplePubSub.java
jsight: @lincolnthree - Any input?

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+        <tag>mq</tag>
+        <tag>mqe</tag>
+        <tag>ibm</tag>
+        <tag>jms</tag>
+    </metadata>
+
+
+    <rules>
+        <!-- MQe*Adapter -->
+        <rule id="websphere-mq-8jyqn">
```
lincolnthree: the rule id's should be updated to meet our naming standards:

-01000
-02000
-03000

...
etc

pom.xml
```diff
@@ -224,7 +231,10 @@
                 <groupId>org.apache.maven.plugins</groupId>
                	<artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <argLine>-Xms512m -Xmx2048m -XX:MaxPermSize=768m -XX:ReservedCodeCacheSize=128m</argLine>
+                    <systemPropertyVariables>
+                        <org.jboss.logging.provider>jdk</org.jboss.logging.provider>
```
lincolnthree: What is this for? Probably should be removed.

pom.xml
```diff
@@ -163,8 +163,15 @@
     <build>
         <resources>
             <resource>
-                <filtering>false</filtering>
                 <directory>rules</directory>
+                <filtering>false</filtering>
+            </resource>
+            <resource>
+                <directory>rules-reviewed/</directory>
```
lincolnthree: What is this for?

rules-reviewed/jboss-eap/websphere/WebSphere-other.windup.xml
```diff
@@ -0,0 +1,72 @@
...
+        <sourceTechnology id="websphere" versionRange="(10,12]"/>
+        <targetTechnology id="eap" versionRange="[6,)"/>
+        <targetTechnology id="java-ee" versionRange="[6,)"/>
+        <tag>reviewed-2015-06-05</tag>
+        <tag>ibm</tag>
+    </metadata>
+
+
+    <rules>
+        <rule id="websphere-other-6bcvk">
```
lincolnthree: the rule id's should be updated to meet our naming standards:

-01000
-02000
-03000

...
etc

rules-reviewed/jboss-eap/websphere/WebSphere-other.windup.xml
```diff
@@ -0,0 +1,72 @@
...
+
+    <metadata>
+        <description>
+            Provides analysis of WebSphere proprietary classes and constructs that may require individual attention when migrating to JBoss EAP 6+.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.2.0.Final"/>
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.2.0.Final"/>
+        </dependencies>
+        <sourceTechnology id="websphere" versionRange="(10,12]"/>
```
lincolnthree: Please remove source versionRange attribute.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+
+    <metadata>
+        <description>
+            Provides analysis of WebSphere JMS proprietary classes and constructs that may require individual attention when migrating to JBoss EAP 6+.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.2.0.Final"/>
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.2.0.Final"/>
+        </dependencies>
+        <sourceTechnology id="websphere" versionRange="(10,12]"/>
```
lincolnthree: Please remove source versionRange.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ JMS adapter." effort="4">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/adapters/package-summary.html"/>
+                </classification>
+                <hint>
+                    <message>
+                        WebSphere Startup Service can be replaced with EJB 3.1 Startup Bean,
```
lincolnthree: This hint needs a solution link.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+            </perform>
+        </rule>
+
+
+
+        <!-- WebSphere JMS MQe -->
+
+        <!--
+            Properties env = new Properties();
+            env.put(
```
lincolnthree: These comments should probably be removed, or they should explain why they are here.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                Context.INITIAL_CONTEXT_FACTORY,
+                "com.ibm.mq.jms.context.WMQInitialContextFactory");
+        -->
+        <rule id="websphere-mq-r322f">
+            <when>
+                <!-- WINDUP-615
```
lincolnthree: This comment is out of date, since we decided that this rule needs to check properties files anyway, right? If so, please remove.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="websphere-mqe-18bicn">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeQueueConnectionFactory"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQe API." description="A more general, and recommended, approach is to use the QueueConnectionFactory class.">
```
lincolnthree: These classifications should not end with a period ".". None of our other classifications do this. Simple fix, just remove the period.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="websphere-mqe-2Aqqrj">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeObjectMessage"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQe API." description="A more general, and recommended, approach is to use the ObjectMessage class.">
```
lincolnthree: I'm not sure that you're using Classification in a way that is really ideal here. It seems like you're using it to add documentation about the actual type reference, but this will be attached to the file itself and will lose context of the reference. This description (and those like it) should be moved into the hint instead.

lincolnthree: Since you link the same MQE documentation in each classification, all of these classifications could easily be replaced by a single rule that matches on the package "com.ibm.mqe"

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
```diff
@@ -0,0 +1,1363 @@
...
+                <hint effort="1">
+                    <message>Replace with Java EE 6 JMS: `javax.jms.XATopicSession`.</message>
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+
+
+        <!-- SIB JMS -->
+        <rule id="websphere-mq-7Apvvd">
```
lincolnthree: These and the following rules need documentation links.

lincolnthree: Every single one of the rules below this line seem to be replaceable with the following single parameterized rule:

         <rule id="websphere-mq-50000">
            <when>
                <javaclass references="com.ibm.msg.client.jms.Jms{type}"/>
            </when>
            <perform>
                <hint effort="1">
                    <message>Migrate to: javax.jms.{type}</message>
                    <link ... />
                </hint>
            </perform>
        </rule>

This should also have doc links.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
jsight: Based upon the content of the classification title and description, I agree that that belongs in a hint. This looks like something that would apply at the line level to me.

A more general rule that attaches a classification to all files that use com.ibm.mqe.jms might make sense as well, and I could see that using classification.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
lincolnthree: Actually it looks like there are a few rules required:

com.ibm.websphere.sib.api.jms -> javax.jms
com.ibm.msg.client.jms -> javax.jms
com.ibm.jms -> javax.jms

lincolnthree: Something like this:

         <rule id="websphere-mq-50000">
            <when>
                <javaclass references="{prefix}{type}"/>
            </when>
            <perform>
                <hint effort="1">
                    <message>Migrate to: javax.jms.{type}</message>
                    <link ... />
                </hint>
            </perform>
            <where param="prefix"><matches pattern="(com.ibm.websphere.sib.api.jms|com.ibm.msg.client.jms|com.ibm.jms)(Jms|JMS)?+"/></where>
        </rule>

This should also have doc links.

rules-reviewed/jboss-eap/websphere/tests/WebSphereMQ.windup.test.xml
lincolnthree: I agree with Jesse, we should ideally reference the ID of the rule that is being tested.

lincolnthree: But what happens when you have more than one test for a rule? E.g. In the scenario where I suggestedd parameterization of the rules in the JMS ruleset in this PR? One parameterized rule could match a ton of things. We might have a few tests just for that one rule. (In that case, we could add a suffix to the test rule ID:

ruleset-id-02000-1
ruleset-id-02000-2
ruleset-id-02000-3

... etc

rules-reviewed/jboss-eap/websphere/tests/samples/WebSphereMQ/com/ibm/examples/AllClasses.java
```diff
@@ -0,0 +1,95 @@
...
+import com.ibm.mq.jms.MQXASession;
+import com.ibm.mq.jms.MQXATopicConnection;
+import com.ibm.mq.jms.MQXATopicConnectionFactory;
+import com.ibm.mq.jms.MQXATopicSession;
+
+public class AllClasses {
+
+    new MQConnection();
+    new MQQueueConnection();
+    new MQConnectionFactory();
```
lincolnthree: Nice :p

rules-reviewed/jboss-eap/websphere/tests/WebSphereMQ.windup.test.xml
jsight: I like ruleset-id-02000-test-n personally. This will also show up in the failure report if there are test failures, which should make finding the offending test-case (and expected original rule) easier.

rules-reviewed/jboss-eap/websphere/tests/samples/other/SimplePubSub.java
lincolnthree: We can't use this. There is no license that I can find in that download, and it requires a click-through license in order to download (that we can't pass on to our users.) This needs to be replaced with our own code.

pom.xml
OndraZizka: It removes the Weld warning about furnace catching all classes' events.

pom.xml
OndraZizka: See the discussion above - it simply makes that directory treated equally by the IDE as rules/ .

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Well I was thinking, I could spend that time on changing IDs, or reviewing other rules. Is this worth the time?

rules-reviewed/jboss-eap/websphere/tests/WebSphereMQ.windup.test.xml
OndraZizka: To be honest, I don't see any difference between looking for [ruleset-id-02000] and the actual hin message that was not found.

OndraZizka: I am creating a XSLT template to generate IDs, and maybe to sync the tests with rules.

rules-reviewed/jboss-eap/websphere/tests/samples/other/SimplePubSub.java
OndraZizka: Working on our own code.

rules-reviewed/jboss-eap/websphere/WebSphere-other.windup.xml
OndraZizka: Just for websphere?

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Is it desirable to only scan .java, .properties and .xml? I could do 3 rules. Or, when we support <or>, use that.

jsight: I think &lt;or&gt; works, but using a parameter with a "when" conditional would be better, IMO.

lincolnthree: Not sure I'm following @jsight  ?

jsight: Pseudocode: 

when(FileContent.filename("{*}.{extension}")
.where("extension").matches("(.java)|(.properties)|(.xml)")

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: That's right. The rules you are refering to remained the same as they were in the original - not assigned to me, so I didn't process them (yet?), but included them not to loose track of them.

rules-reviewed/jboss-eap/websphere/tests/WebSphereMQ.windup.test.xml
OndraZizka: <?xml version="1.0" encoding="UTF-8"?>

<xsl:stylesheet version="1.0"
    xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
    xmlns:wu="http://windup.jboss.org/schema/jboss-ruleset"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
>
    <xsl:output method="xml"/>


    
    <xsl:template match="wu:rule">
        <xsl:copy>
            <xsl:attribute name="id"><xsl:number format="01"/>00</xsl:attribute>
            <xsl:apply-templates select="@*[local-name() != 'id']|node()"/>
        </xsl:copy>
    </xsl:template>

    
    <xsl:template match="wu:ruleset">
        <xsl:copy>
            <xsl:apply-templates select="@*|node()"/>
        </xsl:copy>
    </xsl:template>

    
    <xsl:template match="@*|node()">
        <xsl:copy>
            <xsl:apply-templates select="@*|node()"/>
        </xsl:copy>
    </xsl:template>

</xsl:stylesheet>

jsight: The XSLT will lose any comments in the file, right?

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
lincolnthree: If they weren't assigned to you, then they really shouldn't be included in your PR because that would confuse the next person :) If you feel like doing them and including them, please make sure they are finished. Thanks!

lincolnthree: (That said, I think it would be really easy to finish since the rule I wrote above should (with tests) be all that is needed.)

rules-reviewed/jboss-eap/websphere/tests/samples/WebSphereMQ/com/ibm/examples/AllClasses.java
OndraZizka: Well... :) Is that enough?

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: BTW - if it reduces the number of rules (but not coverage) - is that ok?

jsight: Reducing the number of rules while maintaining (or increasing) coverage is a good thing, IMO.

rules-reviewed/jboss-eap/websphere/WebSphere-other.windup.xml
OndraZizka: Failed to parse XML configuration at: file:/home/ondra/work/Migration/windup-rulesets/rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml due to: The 'sourceTechnology' element must have a non-empty 'versionRange' attribute

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Added

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Added explanation

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka:
```xml
                <filecontent filename="{*}.{ext}" pattern="{*}com.ibm.mq.jms.context.WMQInitialContextFactory{*}"/>
            <where param="ext"><matches pattern="(\.java|\.properties|\.xml)"/></where>
```

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Removed, period.

rules-reviewed/jboss-eap/websphere/WebSphereMQ.windup.xml
OndraZizka: Done as above.

rules-reviewed/jboss-eap/websphere/tests/samples/WebSphereMQ/com/ibm/examples/AllClasses.java
OndraZizka: Fixed a bit.

OndraZizka:
Rebased. The build fails on validation of weblogic/weblogic-webapp.windup.xml against XSD. Otherwise it should be good.

pom.xml
lincolnthree: Ah yes, I noticed that in the IDE. Good.

rules-reviewed/jboss-eap/websphere/tests/websphere-mq.windup.test.xml
```diff
@@ -0,0 +1,500 @@
...
+
+        <metadata>
+            <description>
+                Provides analysis of WebSphere JMS proprietary classes and constructs that may require individual attention when migrating to JBoss EAP 6+.
+            </description>
+            <dependencies>
+                <addon id="org.jboss.windup.rules,windup-rules-javaee,2.2.0.Final"/>
+                <addon id="org.jboss.windup.rules,windup-rules-java,2.2.0.Final"/>
+            </dependencies>
+            <sourceTechnology id="websphere" versionRange="(0,)"/>
```
lincolnthree: probably better just to remove the versionRange. We made it optional in the XSD.

rules-reviewed/jboss-eap/websphere/tests/websphere-mq.windup.test.xml
```diff
@@ -0,0 +1,500 @@
...
+            <tag>mq</tag>
+            <tag>mqe</tag>
+            <tag>ibm</tag>
+            <tag>jms</tag>
+        </metadata>
+
+        <rules>
+            <!-- WebSphere JMS MQe -->
+
+            <!-- Package-matching generic rule - MQ. -->
```
lincolnthree: Is this comment necessary when the rule-id is now specified in the test rule-id?

lincolnthree: Please remove.

rules-reviewed/jboss-eap/websphere/tests/websphere-mq.windup.test.xml
```diff
@@ -0,0 +1,500 @@
...
+                    <not>
+                        <classification-exists classification="WebSphere MQ API"/>
+                    </not>
+                </when>
+                <perform>
+                    <fail message="Classification not found: WebSphere MQ API"/>
+                </perform>
+            </rule>
+
+            <!-- Example code to match:
```
lincolnthree: Is this comment necessary when the rule-id is now specified in the test rule-id? Please remove.

rules-reviewed/jboss-eap/websphere/tests/websphere-mq.windup.test.xml
```diff
@@ -0,0 +1,500 @@
...
+                    <not>
+                        <hint-exists message="For EAP 6, migrate to JBoss HornetQ."/>
+                    </not>
+                </when>
+                <perform>
+                    <fail message="Hint not found: For EAP 6, migrate to JBoss HornetQ."/>
+                </perform>
+            </rule>
+
+            <!-- .scp config files - WINDUPRULE-1 -->
```
lincolnthree: Is this comment necessary when the rule-id is now specified in the test rule-id? Please remove.

rules-reviewed/jboss-eap/websphere/tests/websphere-mq.windup.test.xml
```diff
@@ -0,0 +1,500 @@
...
+                    </not>
+                </when>
+                <perform>
+                    <fail message="Hint not found: A more general, and recommended, approach is to use the Java EE 6 JMS class `javax.jms.XATopicSession`."/>
+                </perform>
+            </rule>
+
+
+
+            <!-- Covers these classes:
```
lincolnthree: Is this comment necessary when the rule-id is now specified in the test rule-id? Please remove.

rules-reviewed/jboss-eap/websphere/tests/websphere-mqe.windup.test.xml
```diff
@@ -0,0 +1,319 @@
+<?xml version="1.0"?>
+<ruletest xmlns="http://windup.jboss.org/schema/jboss-ruleset">
+    <testDataPath>samples/WebSphereMQe</testDataPath>
+    <rulePath>../websphere-mq.windup.xml</rulePath>
+    <ruleset>
+        <metadata></metadata>
+        <rules>
+
+            <!-- MQe*Adapter -->
```
lincolnthree: Please remove comment.

rules-reviewed/jboss-eap/websphere/transformations/xslt/websphere-jboss5-web-xml.xsl
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+
+    <xsl:template match="/">
+        <xsl:text disable-output-escaping='yes'>&lt;!DOCTYPE jboss-web PUBLIC "-//JBoss//DTD Web Application 5.0//EN" "http://www.jboss.org/j2ee/dtd/jboss-web_5_0.dtd"&gt;
```
lincolnthree: Should we still be supporting this translation? Is jboss-web.xml still supported in EAP 6?

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+
+    <metadata>
+        <description>
+            Provides analysis of WebSphere JMS proprietary classes and constructs that may require individual attention when migrating to JBoss EAP 6+.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.2.0.Final"/>
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.2.0.Final"/>
+        </dependencies>
+        <sourceTechnology id="websphere" versionRange="(0,)"/>
```
lincolnthree: Remove source versions.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere MQe JMS adapter" effort="4">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/adapters/package-summary.html"/>
+                </classification>
+                <hint>
+                    <message>
+                        WebSphere MQe JMS adapter can be [deployed to JBoss EAP 6](https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6/html/Administration_and_Configuration_Guide/Deploy_the_WebSphere_MQ_Resource_Adapter.html).
```
lincolnthree: Use full sentences, please.

rules-reviewed/jboss-eap/websphere/transformations/xslt/websphere-jboss5-web-xml.xsl
jsight: Yes, although I don't know if this template is up to date (it looks like it is not).

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+                <javaclass references="com.ibm.mqe.administration.MQe{*}AdminMsg">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ JMS administration API" description="Provides classes used to administer and monitor a Websphere MQ Everyplace queue manager." effort="4">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/adapters/package-summary.html"/>
+                </classification>
+                <hint>
+                    <message>In JBoss EAP 6, the JMS destinations can be controlled with JBoss CLI.</message>
```
lincolnthree: Okay, that's good, but we need an explanation of the problem scenario that was found.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+            <when>
+                <javaclass references="com.ibm.mqe.adapters.MQe{*}Adapter">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere MQe JMS adapter" effort="4">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/adapters/package-summary.html"/>
+                </classification>
+                <hint>
```
jsight: These need a severity attribute, and I think a title would be helpful as well.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+
+        <!-- MQe*AdminMsg -->
+        <rule id="websphere-mq-02000">
+            <when>
+                <javaclass references="com.ibm.mqe.administration.MQe{*}AdminMsg">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ JMS administration API" description="Provides classes used to administer and monitor a Websphere MQ Everyplace queue manager." effort="4">
```
lincolnthree: This classification should really be split into a separate rule that matches on "com.ibm.mqe.administration", I think.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+                </or>
+                -->
+                <filecontent filename="{*}.{ext}" pattern="{*}com.ibm.mq.jms.context.WMQInitialContextFactory{*}"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ JMS InitialContextFactory">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/jms/MQeQueueConnection.html"/>
+                </classification>
+                <hint effort="1">
+                    <message>For EAP 6, migrate to JBoss HornetQ.</message>
```
lincolnthree: This hint needs to be a little more verbose, for example:

"The IBM JMS InitialContextFactory is used to do ABC and XYZ. The equivalent functionality needs to be configured on JBoss EAP 6 using HornetQ.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+                    <filecontent filename="{*}.xml" pattern="{*}&quot;com.ibm.mq.jms.context.WMQInitialContextFactory&quot;{*}"/>
+                </or>
+                -->
+                <filecontent filename="{*}.{ext}" pattern="{*}com.ibm.mq.jms.context.WMQInitialContextFactory{*}"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ JMS InitialContextFactory">
+                    <link description="Javadoc" href="http://www-01.ibm.com/support/knowledgecenter/SSVHEW_6.2.0/com.ibm.mqe.javadoc/java/com/ibm/mqe/jms/MQeQueueConnection.html"/>
+                </classification>
+                <hint effort="1">
```
lincolnthree: Is this really a trivial change? Seems like this might be a "3" - "Complex migration effort with documented solution"

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+        <!-- .scp config files - WINDUPRULE-1 -->
+        <rule id="websphere-mq-04000">
+            <when>
+                <filecontent filename="{*}.scp" pattern="{*}"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ JMS configuration file">
+                </classification>
+                <hint effort="3">
+                    <message>The recommended method of configuring messaging in JBoss EAP 6 is in either the Management Console or Management CLI.</message>
```
lincolnthree: Good, but this hint needs to be a bit more verbose:

"WebSphere `.scp` files are used to ABC and XYZ. these need to be migrated to JBoss EAP 6 JMS using the Management Console or Management CLI."

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+        </rule>
+
+
+
+        <rule id="websphere-mq-05000">
+            <when>
+                <javaclass references="com.ibm.msg.client.wmq.WMQConstants"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ Everyplace JMS constants">
```
lincolnthree: Do we really need a classification for every kind of file? Seems like a single IBM messaging classification at the package level `com.ibm.msg.client`, etc, would be more appropriate.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+                </hint>
+            </perform>
+        </rule>
+
+        <rule id="websphere-mq-06000">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeJMSAuthenticator"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQ Everyplace JMS authentication interface">
```
lincolnthree: Again, do we really need a classification on every single type reference? This is going to blow up the report. Classifications should be succinct. "WebSphere Proprietary JMS", or something like that at the package level.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+                </hint>
+            </perform>
+        </rule>
+
+        <rule id="websphere-mq-07000">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeJMSMsgFieldNames"/>
+            </when>
+            <perform>
+                <classification title="WebSphere MQe constants" description="Defines a number of String constants that are used to name fields when a JMS message is mapped to an MQeMsgObject.">
```
lincolnthree: Seems like a hint would be more appropriate here?

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+        </rule>
+
+
+        <rule id="websphere-mq-08000">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeConnection"/>
+            </when>
+            <perform>
+                <hint effort="1">
+                    <message>A more general, and recommended, approach is to use the Java EE 6 JMS class `javax.jms.Connection`.</message>
```
lincolnthree: This hint needs to be a little clearer. First, the approach is required, not recommended. Second, an explanation of the problem is also useful to help people understand what is going on. Such as my earlier comments on other hints that need a little more explanation.

lincolnthree: Same for the other hints in this file (below).

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+            <perform>
+                <hint effort="1">
+                    <message>A more general, and recommended, approach is to use the Java EE 6 JMS class `javax.jms.TemporaryQueue`.</message>
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="websphere-mq-20000">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeBytesMessage"/>
```
lincolnthree: Some of these rules can be combined into a single parameterized rule, like this: 
https://github.com/windup/windup-rulesets/pull/35/files#diff-8b8691b4248f1671a4771ceb9074a33aR116

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,846 @@
...
+            com.ibm.msg.client.jms.JmsXATopicConnectionFactory
+            com.ibm.msg.client.jms.JmsXATopicSession
+            com.ibm.jms.JMSBytesMessage
+            com.ibm.jms.JMSMapMessage
+            com.ibm.jms.JMSMessage
+            com.ibm.jms.JMSObjectMessage
+            com.ibm.jms.JMSStreamMessage
+            com.ibm.jms.JMSTextMessage
+        -->
+        <rule id="websphere-mq-62000">
```
lincolnthree: Good, but a little more explanation would be nice here so that people are oriented and understand what is going on when they read the hint. Again, several complete sentences would be nice.

rules-reviewed/jboss-eap/websphere/websphere-other.windup.xml
```diff
@@ -0,0 +1,71 @@
...
+        <rule id="websphere-other-01000">
+            <when>
+                <javaclass references="com.ibm.websphere.asynchbeans.Work">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere Asynchronous Work"/>
+                <hint effort="3">
+                    <message>Migrate to JCA Work Manager.</message>
```
lincolnthree: More verbosity please.

rules-reviewed/jboss-eap/websphere/websphere-other.windup.xml
```diff
@@ -0,0 +1,71 @@
...
+        <!-- Startup Service -->
+        <rule id="websphere-other-02000">
+            <when>
+                <javaclass references="com.ibm.websphere.startupservice.{*}StartUp{*}">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere Startup Service" effort="4"/>
+                <hint effort="1">
```
lincolnthree: It appears that all of your hints need to be updated with MANDATORY/OPTIONAL as appropriate. I don't recall seeing any of them in this PR.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Even in case there is nothing else than classes matching that pattern?

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Added MANDATORY to all.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Done, to the extent of what I found in the guide.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Changed

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Done

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: That's what we had before, I am just continuing in that sense. I agree that having general rule would be better.

OndraZizka: Added.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Changed.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Yes. Changed.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: That's right, but these classes map 1:1 to JMS API. Should I copy the Javadoc for them?

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Some of these rules were combined.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Expanded a bit.

rules-reviewed/jboss-eap/websphere/websphere-other.windup.xml
OndraZizka: Added better description.

rules-reviewed/jboss-eap/websphere/websphere-other.windup.xml
OndraZizka: Right, added. Most of them are Mandatory.  (it's capitalized in Rules - properties)

pom.xml
lincolnthree: Ok. I hadn't noticed that.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,363 @@
...
+            <when>
+                <javaclass references="{prefix}{type}"/>
+            </when>
+            <perform>
+                <hint severity="Mandatory" effort="1">
+                    <message>
+                        `{prefix}{type}` is a proprietary WebSphere MQ implementation of standard Java EE JMS interface.
```
lincolnthree: Excellent! Now we're talking!

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,363 @@
...
+                        To migrate to JBoss EAP 6, use the Java EE 6 JMS class `javax.jms.{type}`.
+                    </message>
+                </hint>
+            </perform>
+            <where param="prefix"><matches pattern="com\.ibm(\.websphere\.sib\.api|\.msg.client)?\.jms\.(Jms|JMS)[^\.]+"/></where>
```
lincolnthree: com.ibm doesn't need to be in the prefix pattern, does it? Could just be up in the javaclass:

`<javaclass references="com.ibm{prefix}{type}"/>`

That would simplify the pattern to:

`pattern="(\.websphere\.sib\.api|\.msg.client)?\.jms\.(Jms|JMS)[^\.]+"`

lincolnthree: Also, I don't believe you need to escape `.` in character classes:

`[^\.]` should be `[^.]`, but I don't think `\` will really interfere here because it's not a valid java name.

lincolnthree: But this is all fine. Just enhancements, not required.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,363 @@
...
+        -->
+        <rule id="websphere-mq-62000">
+            <when>
+                <javaclass references="{prefix}{type}"/>
+            </when>
+            <perform>
+                <hint severity="Mandatory" effort="1">
+                    <message>
+                        `{prefix}{type}` is a proprietary WebSphere MQ implementation of standard Java EE JMS interface.
+                        To migrate to JBoss EAP 6, use the Java EE 6 JMS class `javax.jms.{type}`.
```
lincolnthree: This hint does still need a link, though. Probably to the Java EE docs and also to the WebSphere docs (not the specific class, just the JMS JavaEE doc: 

https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,363 @@
...
+        </rule>
+
+
+        <rule id="websphere-mq-11000">
+            <when>
+                <javaclass references="com.ibm.mqe.jms.MQeJNDIConnectionFactory"/>
+            </when>
+            <perform>
+                <hint severity="Mandatory" effort="1">
+                    <message>To migrate to EAP 6, use the Java EE 6 JMS class `javax.jms.ConnectionFactory`.</message>
```
lincolnthree: These hints need to be expanded a little bit (like you did with the others.)

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,363 @@
...
+                <javaclass references="com.ibm.mqe.jms.MQeJNDIQueueConnectionFactory"/>
+            </when>
+            <perform>
+                <hint severity="Mandatory" effort="1">
+                    <message>To migrate to EAP 6, use the Java EE 6 JMS class `javax.jms.ConnectionFactory`.</message>
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="websphere-mq-16000">
```
lincolnthree: What happened to 13000-15000? ;)

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
```diff
@@ -0,0 +1,363 @@
...
+            <perform>
+                <hint severity="Mandatory" effort="1">
+                    <message>To migrate to EAP 6, use the Java EE 6 JMS class `javax.jms.{type}`.</message>
+                    <link description="Java EE 6 JMS Tutorial" href="https://docs.oracle.com/javaee/6/tutorial/doc/bnceh.html"/>
+                </hint>
+            </perform>
+        </rule>
+
+
+        <!-- WebSphere JMS MQ -->
```
lincolnthree: Should this still be 61000 when the last one is 18000? :) (This is a good problem to have. Means we reduced rule verbosity!)

rules-reviewed/jboss-eap/websphere/websphere-other.windup.xml
```diff
@@ -0,0 +1,83 @@
...
+                <javaclass references="com.ibm.websphere.asynchbeans.Work">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <classification title="WebSphere Asynchronous Work"/>
+                <hint severity="Mandatory" effort="3">
+                    <message>
+                        WebSphere Asynchronous Work is a Work Manager, whose purpose is to
+                        allow the user utilizing threads while letting the container manage them.
```
lincolnthree: Excellent.

lincolnthree:
The commits in this PR need to be squashed together into a single commit.

pom.xml
OndraZizka: A lot of these:

 2015-06-15 18:35:27 WARN   org.jboss.weld.event.ExtensionObserverMethodImpl checkRequiredTypeAnnotations() WELD-000411: Observer method [BackedAnnotatedMethod] public org.jboss.forge.furnace.container.cdi.impl.ContainerServiceExtension.processRemoteServiceTypes(@Observes ProcessAnnotatedType<Object>) receives events for all annotated types. Consider restricting events using @WithAnnotations or a generic type with bounds.

A lot means, 10's

OndraZizka: Could you perhaps share your JDK logging config - if you have any? Maybe it's me doing it wrong.

jsight: Without this patch applied, I see that line about 40 times. With that patch applied, I see that line about 40 times. I also do not understand how this property could accect that line?

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: The rules numbering will not always follow 1000's step. Does it really matter if we have that state now?

OndraZizka: IoW - should I renumber all the rules again, and the tests too?

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: I  don't remember. I kept an eye not to loose the logic though.

OndraZizka:
I will squash it once it is agreed to be merged, otherwise I have a hard time explaining the changes.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Renumbered.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Done

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Links added.

rules-reviewed/jboss-eap/websphere/websphere-mq.windup.xml
OndraZizka: Applied

OndraZizka: The [^..] part should not be there actually. com.ibm done so.

lincolnthree:
Merged as part of https://github.com/windup/windup-rulesets/commit/0eb98f7b4f14e0054eea1c458e7e0fceadeb8b6e

PR #26, opened 2015-06-02
WINDUP-612  Validate all XML rules in windup-rulesets during build.

OndraZizka:


src/test/java/org/jboss/windup/rules/tests/LineitemExistsHandler.java
```diff
@@ -29,7 +29,7 @@
  * @author mnovotny
  *
  */
-@NamespaceElementHandler(elementName = LineitemExistsHandler.ELEMENT_NAME, namespace = "http://windup.jboss.org/v1/xml")
+@NamespaceElementHandler(elementName = LineitemExistsHandler.ELEMENT_NAME, namespace = "http://windup.jboss.org/schema/jboss-ruleset")
```
lincolnthree: We will probably want a separate namespace for the test handlers I think.

PR #28, opened 2015-06-03
Move WebLogic proprietary servlet annotation rule from quickstarts to…

sgilda:
… the core ruleset


jsight:
Continuing in https://github.com/windup/windup-rulesets/pull/33

PR #30, opened 2015-06-04
Spring test initialization

mbriskar:
I was suspicious that nested rules do not work so I created test. But they work. However, why not to merge the test then? :)


lincolnthree:
Nice job, matej.

lincolnthree:
Merged.

PR #37, opened 2015-06-10
WINDUP-633: Remove conditions that were moved to windup/windup

mbriskar:


mbriskar:
this is probably outdated and should be closed. Or?

jsight:
Yes, I think this can be closed for the moment. Although we do need to move this testing framework into windup core at some point.

PR #40, opened 2015-06-15
Make the -DrunTestsMatching actually work

OndraZizka:
Without letting Maven to pass it to surefire, mvn -DrunTestsMatching does nothing.


jsight:
It works here without this and the documention doesn't seem to imply that this should be necessary.

OndraZizka:
You're right. It was a typo in my config. Sorry for the noise.

PR #43, opened 2015-06-17
WINDUP-654 - Fixed package mappings to be literal as opposed to regex…

bradsdavis:
…, which was overkill.


lincolnthree:
Yep. Looks like these aren't regexes. They probably could be improved to use regex at some point.

PR #62, opened 2015-08-06
https://issues.jboss.org/browse/WINDUP-710 - adding XSD for ruletest …

mareknovotny:
…and updated *.windup.test.xml files to it and added validation to schema in the build.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -212,6 +212,13 @@
                             <dir>rules-reviewed</dir>
                             <systemId>${project.build.directory}/xsd/schema/windup-jboss-ruleset.xsd</systemId>
                         </validationSet>
+                        <validationSet>
+                            <includes>
+                                <include>**/*.windup.test.xml</include>
+                            </includes>
+                            <dir>rules-reviewed</dir>                            
+                            <systemId>schema/windup-jboss-ruletest.xsd</systemId>
+                        </validationSet>                        
                     </validationSets>
                 </configuration>
             </plugin>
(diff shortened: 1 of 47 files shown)


mareknovotny:
This PR is like a preview how it should work, we can change XSD definition(s) on request as I grabbed mostly what is in existing *.windup.test.xml files so far.

mareknovotny:
I rebased the PR commit not to conflict with master branch

mareknovotny:
rebased and updated to current status quo

jsight:
This does not build as the ruleset namespace is not the same as the test namespace. So none of the test rulesets parse correctly.

mareknovotny:
Do you mean by different namespace this https://github.com/mareknovotny/windup-rulesets/commit/f5e043ef5eb61034f782294aeeaa639827dc167a#diff-38c76d42b2361873851466c4de3a300dR45 ?

I will rebase to current status and update new rules to have ruletest XSD too.

jsight:
When I ran the tests, they failed on every test due to a NPE as they didn't
load any rules.

mareknovotny:
interesting ... I rebased the branch and i got the validation errors, but that is due some wrong merges, as I can't see there correct namespace in declaration. I will go through them and do update to correct one.

mareknovotny:
I got what you meant, different namespace in element handlers breaks the loading of test rules. I haven't considered the reusing them in xsd breaks testing rules. I will think about it.

mareknovotny:
I am closing this as I will update rather https://github.com/windup/windup/blob/master/config-xml/schema/windup-jboss-ruleset.xsd to have missing windup test elements and will create new PR for updating existing windup rulesets with it.

PR #65, opened 2015-08-07
#105 QA - SonicESB to Camel

mareknovotny:


mareknovotny:
I commented failing test sonic-esb-03000-test as I don't quite understand why that rule is not firing. I would appreciate a help

jsight:
The error is this:

withEffort(4).withTags([Spring, Apache Camel, SonicESB]))).withId("sonic-esb-03000")
Character at starting position is escaped, and cannot be used in capturing a group.
java.lang.IllegalArgumentException: Character at starting position is escaped, and cannot be used in capturing a group.


This appears to be due to two things:

1. The { needs to be escaped
2. The \{ would fail anyway due to what looks like (to me) a rewrite bug.

I will send an email to you with further details. A few other hints will likely need to be updated in this PR with the escapes as well.

lincolnthree:
Looking at this.

lincolnthree:
Okay. I figured out the bug. It's a problem with RegexParameterizedPatternBuilder in rewrite. Apparently i've been doing a good job of fixing and testing the parser, but not so much with the builder.

lincolnthree:
Fixed and merged.

PR #66, opened 2015-08-17
QA'd some of the uncategorised rules

mbriskar:
I removed from rules/uncategorized of those that were migrated. I didn't do all of them because migrating this was not yet decided. So this is more like a bonus for this sprint in which I migrated those that I subjectively think have the biggest added value.


lincolnthree:
Hey Matej, this looks great! One question. Do you think we should leave these uncategorized? I'm thinking we should probably figure out how to categorize them as part of their inclusion. Merging this for now, but I think we should figure that out.

mbriskar:
I agree, that's also a reason why it is catch-all only

jsight:
I think "universal" might be a better name than uncategorized. These rules apply regardless of source platform.

PR #71, opened 2015-09-08
WINDUPRULE-78 added title for websphere-xml-05000 and created test fo…

mareknovotny:
…r hint existence and noticed too test rules have got the same ids so added -test to all of them.


mareknovotny:
I rebased the PR from master

PR #83, opened 2015-10-01
Removing "rule generic-catchall-00800"

Maarc:
It is generating only false positives and its content is already fully covered by our very nicht EJB/JPA/Server resources reports.

### DIFF
--- a/rules-reviewed/jboss-eap/uncategorized/generic-catchall.windup.xml
+++ b/rules-reviewed/jboss-eap/uncategorized/generic-catchall.windup.xml
@@ -235,22 +235,6 @@
         </rule>
         <rule id="generic-catchall-00800">
             <when>
-                <javaclass references="javax.{subpackage}.{*}">
-                </javaclass>
-            </when>
-            <perform>
-               <classification effort="0" severity="optional" title="Java EE API"/>
-               <hint effort="1" title="{subpackage} type reference" severity="optional">
-                    <message>Java EE API {subpackage} type reference found. No specific details available.</message>
-                    <tag>catchall</tag>
-               </hint>
-            </perform>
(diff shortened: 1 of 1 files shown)


jsight:
@mareknovotny - Can you take a look at this one?

mareknovotny:
@jsight sure.

Maarc:
@mareknovotny 

As described in the pull request, this rule is generating an important amount of false positives. 

Considered that we already have dedicated EJB / JPA  ... reports describing what the application is doing, there is no point at keeping this rule. We do not need it any more. It is only generating "noise".

mareknovotny:
@Maarc I got your point and on that problem I noticed that my original change is missing the important thing to add hints/classifications only if there is not already hint added - this should be in all catchall rules to show up only if there is not any specifics about the matched group.

This change should resolve your concerns with false positive, please try it from https://github.com/windup/windup-rulesets/pull/87.

Maarc:
Hi Marek,

#87 does not solve my "problem". I just want to remove this one rule because it is already covers by other reports (dedicated EJB / JPA reports) and only produces noise.

I still think, that we should comment out this rule.

best regards,

Marc

jsight:
@rdruss - I agree with @Maarc. What do you think? And is there anything we need to do to document or get approval for this change?

mareknovotny:
@Maarc @jsight @rdruss I did that in https://github.com/mareknovotny/windup-rulesets/commit/cb3ea8cb7b620fc3861e7b825993413c2e9f4ed5

jsight:
@mareknovotny incorporated this into https://github.com/windup/windup-rulesets/pull/87 and I have merged that one.

PR #102, opened 2015-12-09
WINDUPRULE-101: Rule for t3:// and environment.put("java.naming.facto…

jsight:
https://issues.jboss.org/browse/WINDUPRULE-101


rdruss:
I do not believe that the T3 protocol is limited to JNDI queries as it's an RMI implementation

jsight:
Yes, that's the bad thing about this rule. It _can_ be used in a lot of different contexts. How do we provide useful migration assistance to the user based upon the other uses of a bare string with this protocol specified?

PR #108, opened 2016-02-01
WINDUPRULE-111 and WINDUPRULE-110

mbriskar:


Ladicek:
Change the `eap6-xml-02000` rule in `eap6-xml.windup.xml` too? Please? :-)

mbriskar:
sure :)

Ladicek:
It still says that "There is no workaround currently." The message in the new rule "JBoss Web was replaced by Undertow, which does not contain support for the Valve functionality. It may be needed to migrate to Undertow handlers" is better.

mbriskar:
Done. What about the documentation?

Ladicek:
What kind of documentation do you think of?

mbriskar:
About how to migrate to undertow handlers from valves

Ladicek:
Right. We have something in EAP 7 Migration Guide, if you want to see that, ping me on IRC :-)

Ladicek:
LGTM. The documentation links point to a section of documentation that speaks about subsystem migration, but the part of the documentation that speaks about application migration is plain wrong, so this will be better for now.

Ladicek:
Thanks for the other commit. Looks good to me.

In the meantime, I fixed https://docs.jboss.org/author/display/WFLY10/How+do+I+migrate+my+application+from+AS7+to+WildFly#HowdoImigratemyapplicationfromAS7toWildFly-JBossWebValves so that it now contains some actually useful info. Can we link to that instead, in the rules for valves?

mbriskar:
Ok I have referenced that link. The only thing that I am not that much satisfied with is xpath "//valve". This will be slow and will run on every xml file that will be registered

Ladicek:
It's enough to look for `<valve>` only in `jboss-web.xml`.

mbriskar:
oh good point

Ladicek:
One other thing -- AFAIK, you don't really have to look for `//valve`, it's enough to look for `/jboss-web/valve`.

mareknovotny:
@mbriskar have you seen previous comment about 

> it's enough to look for `/jboss-web/valve`.
?

mbriskar:
I have changed the //valve to jboss-web/valve

PR #110, opened 2016-02-02
WINDUPRULE-112: Insufficient rule for jboss-deployment-structure.xml

mbriskar:


Ladicek:
Ha, I like how this gets shorter while also getting more correct :-)

mareknovotny:
+1

PR #112, opened 2016-02-04
WINDUP-911 take only *.windup.xml files for built jar artifact

mareknovotny:
https://issues.jboss.org/browse/WINDUP-911

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -223,6 +223,17 @@
                     <argLine>-Xms512m -Xmx2048m -XX:MaxPermSize=768m -XX:ReservedCodeCacheSize=128m</argLine>
                	</configuration>
             </plugin>
+            <plugin>
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-jar-plugin</artifactId>
+               <configuration>
+                  <includes>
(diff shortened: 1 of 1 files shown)


jsight:
Won't this also exclude XSLT files?

mareknovotny:
yes, it will, haven't thought about them :(, I will add an inclusion for them

jsight:
This is also filtering out API jars that are used for class reachability analysis (when that feature is enabled).

mareknovotny:
@jsight  do you mean the java API? I will add them too then.

mareknovotny:
I got it now, I didn't know they are here in rulesets in jboss-eap/api-jars/*.jars

OndraZizka:
I have tried this using
 --input /home/ondra/work/Migration/TestApps/_apps/jee-example-app-1.0.0-unparsable.ear --output /tmp/Windup/resJee  --overwrite --batchMode --debugx --keepWorkDirs --target eap --enableClassNotFoundAnalysis
and it worked fine.

OndraZizka:
But now I checked and the artifact still has 17 MB.

PR #116, opened 2016-02-11
URL Correction for JMS Configuration documentation

randythomas:
…the following:

https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6.4/html-single/Administration_and_Configuration_Guide/index.html#sect-Configuration

This URL leads to configuration information for Hibernate. i.e. it is the wrong URL. The correct URL is the following:

https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6.4/html-single/Administration_and_Configuration_Guide/index.html#sect-Configuration1

This URL goes to the JMS Configuration information section of the document, which is what is required. The "1" at the end of the URL is the only difference the two URLs


jsight:
Nice! Thanks for the fix @randythomas!

PR #126, opened 2016-03-11
WINDUPRULE-132 fixing issues for Hibernate 5 rules

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-132


rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -10,9 +10,10 @@
             <addon id="org.jboss.windup.rules,windup-rules-java,2.4.0.Final" />
         </dependencies>
         <sourceTechnology id="hibernate" versionRange="[4,)" />
+        <sourceTechnology id="eap" versionRange="[4,7)" />
```
Ladicek: Can we add source technology `eap` also to `hibernate4-xml.windup.xml`?

Also, I don't think this applies to EAP 4 and 5, it should be `<sourceTechnology id="eap" versionRange="[6,7)" />`.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +43,42 @@
...
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+            <where param="idclass">
+                <matches pattern="(TableGenerator|TableHiLoGenerator|SequenceGenerator|SequenceIdentityGenerator|SequenceHiLoGenerator)" />
+            </where>
+        </rule>
+        <rule id="hibernate4-00003">
+            <when>
+                <xmlfile matches="//generator[@class='hilo'] | //generator[@class='sequence-identity']" in="{*}.hbm.xml"/>
```
Ladicek: In addition to `hilo` and `sequence-identity`, there are also `sequence` and `seqhilo`.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +43,42 @@
...
+            </when>
+            <perform>
+                <hint title="Identifier generator was removed in Hibernate 5" effort="1" severity="mandatory">
+                    <message>Use `sequence` generator instead.</message>
```
Ladicek: Code name of `SequenceStyleGenerator` is `enhanced-sequence`.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
                 <matches pattern="(TableGenerator|TableHiLoGenerator|SequenceGenerator|SequenceIdentityGenerator|SequenceHiLoGenerator)" />
             </where>
         </rule>
+        <rule id="hibernate4-00029">
+            <when>
+                <javaclass references="org.hibernate.type.descriptor.sql.SqlTypeDescriptor">
```
Ladicek: I'm not sure this will work correctly. The text refers to "extending built-in SqlTypeDescriptors" -- is this rule going to match them?

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
Ladicek: Aha, I see that `sequence` now also points to `SequenceStyleGenerator`.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
Ladicek: As I note below, `sequence` now refers to `SequenceStyleGenerator`. So maybe we don't have to catch it at all, maybe it should have a different migration warning (effort 0, just to inform users that it changed meaning).

mareknovotny: Steve Ebersole wrote only about _hilo_ and _sequence-identity_ removal. So even others can be deprecated they should work or they are replaced on class level detail. Not sure with emphasizing them ;)

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: np. I will add/change that.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: right there is commented _location_ element. I played with this and it was somehow failing but it should be caught by that _INHERITANCE_ type. good catch!

Ladicek: I think the rule will catch inheritance from `SqlTypeDescriptor`, but it won't catch inheritance from the built-in `SqlTypeDescriptor`s, which themselves inherit from `SqlTypeDescriptor`, no?

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
Ladicek: `hilo` has indeed been removed.

`sequence-identity` is still there and corresponds to `SequenceIdentityGenerator` which is deprecated.

`seqhilo` is the same case: still present, corresponds to `SequenceHiLoGenerator`, which is also deprecated.

See:
old -- https://github.com/hibernate/hibernate-orm/blob/4.2/hibernate-core/src/main/java/org/hibernate/id/factory/internal/DefaultIdentifierGeneratorFactory.java
new -- https://github.com/hibernate/hibernate-orm/blob/5.0/hibernate-core/src/main/java/org/hibernate/id/factory/internal/DefaultIdentifierGeneratorFactory.java

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: see http://windup.github.io/windup/docs/latest/javadoc/org/jboss/windup/ast/java/data/TypeReferenceLocation.html#INHERITANCE
it says "works transitively" I would assume it should work. Or not @jsight, @OndraZizka ?

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
mareknovotny: ok, that changes the thing, I keep there _hilo_ only.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
Ladicek: I guess that it only works transitively if the entire inheritance transitive closure is available for Windup to analyze. When migrating an EAP application, I don't expect Hibernate core to be available. But maybe I'm misunderstanding here.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
Ladicek: Well for classes, we also mark those that are still available but deprecated. I'd expect that for code names, we'd do the same.

mareknovotny: ok, I will add deprecated list too - 'sequence-identity', 'seqhilo', 'seqhilo', 'enhanced-table

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: That is true. So that rule is hard to design and just reference without type location is not enough. I need to discuss this with @jsight

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +44,44 @@
...
+                </hint>
+            </perform>
+            <where param="idclass">
+                <matches pattern="(TableGenerator|TableHiLoGenerator|SequenceGenerator|SequenceIdentityGenerator|SequenceHiLoGenerator)" />
+            </where>
+        </rule>
+        <rule id="hibernate4-00003">
+            <when>
+                <xmlfile 
+                 matches="//generator[@class='hilo'] | //generator[@class='sequence-identity'] | //generator[@class='seqhilo']| //generator[@class='seqhilo']| //generator[@class='enhanced-table']"
```
Ladicek: `seqhilo` twice

Ladicek: Is `enhanced-table` also deprecated? I didn't think so.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
jsight: Yes, it needs them on the library path in some way in order for it to match transitively.

mareknovotny: so even I tried a hack with targeting eap, it doesn't find it :(

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -11,8 +11,10 @@
             <addon id="org.jboss.windup.rules,windup-rules-java,2.4.0.Final" />
         </dependencies>
         <sourceTechnology id="hibernate" versionRange="[4,5)" />
+        <sourceTechnology id="eap" versionRange="[6,7)" />
         <targetTechnology id="hibernate5" versionRange="[5,)" />
         <targetTechnology id="eap7" versionRange="[7,)" />
+        <targetTechnology id="eap" versionRange="[7,)" />
```
Ladicek: Is this OK? I don't think we've been using target technology of `eap` for EAP 7, only `eap7`.

I mean, if we were able to properly use the version ranges when invoking Windup (like `windup --source eap:6 --target eap:7`), that would be awesome, but we're not there yet, no?

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +45,44 @@
...
+            </where>
+        </rule>
+        <rule id="hibernate4-00003">
+            <when>
+                <xmlfile 
+                 matches="//generator[@class='hilo'] | //generator[@class='sequence-identity'] | //generator[@class='seqhilo']"
+                 in="{*}.hbm.xml"/>
+            </when>
+            <perform>
+                <hint title="Identifier generator was removed in Hibernate 5" effort="1" severity="mandatory">
```
Ladicek: This can't be made to show the exact identifier, like the rule above, can it? I'd prefer it to read `Generator XXX was removed/deprecated in Hibernate 5`.

If not, then at least `This generator was removed/deprecated in Hibernate 5`.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+            </perform>
+        </rule>
+        <rule id="hibernate4-00033">
+            <when>
+                <javaclass references="org.hibernate.integrator.spi.Integrator">
+                    <location>IMPLEMENTS_TYPE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Interface org.hibernate.integrator.spi.Integrator were changed in Hibernate 5" effort="1" severity="optional">
```
Ladicek: Interface _was_ changed

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+        </rule>
+        <rule id="hibernate4-00036">
+            <when>
+                <javaclass references="org.hibernate.id.PersistentIdentifierGenerator">
+                    <location>IMPLEMENTS_TYPE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Interface org.hibernate.id.Configurable changed method parameter in configure method in Hibernate 5" effort="1" severity="mandatory">
+                    <message>Hibernate 5 introduces `org.hibernate.boot.model.relational.ExportableProducer` interface which will effect any of `org.hibernate.id.PersistentIdentifierGenerator` implementations.
```
Ladicek: which will effect any of ... implementations -> which will affect all ... implementations

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+                     <link
+                        href="https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/id/Configurable.html#configure-org.hibernate.type.Type-java.util.Properties-org.hibernate.service.ServiceRegistry-"
+                        title="Hibernate 5 Configurable API" />
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hibernate4-00036">
+            <when>
+                <javaclass references="org.hibernate.id.PersistentIdentifierGenerator">
```
Ladicek: Should this instead look for implementations of `org.hibernate.id.Configurable`?

Ladicek: I mean, all reasonable `PersistentIdentifierGenerator`s probably implement `Configurable`, but they have to do that explicitly.

Ladicek: Oh, I get it now. This rule mixes 2 things together:

- The `PersistentIdentifierGenerator` interface now extends `ExportableProducer`, so all `PersistentIdentifierGenerator` implementations have to implement a new method.
- The `Configurable` interface changed a type of one parameter of the `configure` method.

I think this rule should be split.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hibernate4-00030">
+            <when>
+                <javaclass references="org.hibernate.hql.spi.{type}" />
+            </when>
+            <perform>
+                <hint title="Class org.hibernate.hql.spi.{type} were moved in Hibernate 5" effort="1" severity="mandatory">
```
Ladicek: _was_ moved

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+            <where param="type">
+                <matches pattern="(MultiTableBulkIdStrategy|MultiTableBulkIdStrategy.DeleteHandler|MultiTableBulkIdStrategy.UpdateHandler|AbstractTableBasedBulkIdHandler|AbstractTableBasedBulkIdHandler.ProcessedWhereClause|TableBasedDeleteHandlerImpl|TableBasedUpdateHandlerImpl)" />
+            </where>
+        </rule>
+        <rule id="hibernate4-00031">
+            <when>
+                <javaclass references="org.hibernate.hql.spi.PersistentTableBulkIdStrategy" />
+            </when>
+            <perform>
+                <hint title="Class org.hibernate.hql.spi.PersistentTableBulkIdStrategy were moved in Hibernate 5" effort="1" severity="mandatory">
```
Ladicek: _was_ moved

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hibernate4-00032">
+            <when>
+                <javaclass references="org.hibernate.hql.spi.TemporaryTableBulkIdStrategy" />
+            </when>
+            <perform>
+                <hint title="Class org.hibernate.hql.spi.TemporaryTableBulkIdStrategy were replaced in Hibernate 5" effort="1" severity="mandatory">
```
Ladicek: _was_ replaced

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+            <where param="type">
+                <matches pattern="(MultiTableBulkIdStrategy|MultiTableBulkIdStrategy.DeleteHandler|MultiTableBulkIdStrategy.UpdateHandler|AbstractTableBasedBulkIdHandler|AbstractTableBasedBulkIdHandler.ProcessedWhereClause|TableBasedDeleteHandlerImpl|TableBasedUpdateHandlerImpl)" />
+            </where>
+        </rule>
+        <rule id="hibernate4-00031">
+            <when>
+                <javaclass references="org.hibernate.hql.spi.PersistentTableBulkIdStrategy" />
```
Ladicek: This class can also be referenced in configuration under key `hibernate.hql.bulk_id_strategy`. Should have a rule for that?

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+                    <link
+                        href="https://github.com/hibernate/hibernate-orm/blob/5.0/migration-guide.adoc#misc"
+                        title="Hibernate 5 Misc changes" />
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hibernate4-00032">
+            <when>
+                <javaclass references="org.hibernate.hql.spi.TemporaryTableBulkIdStrategy" />
```
Ladicek: This class can also be referenced in configuration under key `hibernate.hql.bulk_id_strategy`. Should have a rule for that?

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,160 @@
...
+            </perform>
+        </rule>
+        <rule id="hibernate4-00036">
+            <when>
+                <javaclass references="org.hibernate.id.PersistentIdentifierGenerator">
+                    <location>IMPLEMENTS_TYPE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Interface org.hibernate.id.Configurable changed method parameter in configure method in Hibernate 5" effort="1" severity="mandatory">
```
Ladicek: The hint is a copy/paste from previous rule. For this rule, it should say something like "since Hibernate 5, PersistentIdentifierGenerator extends ExportableProducer"

(I deleted my previous comments to this rule because they were misleading. This one finally makes sense.)

rules-reviewed/eap7/eap6/hibernate4.windup.xml
jsight: @mareknovotny - https://github.com/windup/windup/pull/858

This also requires that the rule use "implements_type" instead of inherits (since the sqlTypeDescriptor is an interface). And it still doesn't solve the issue of the correct jars not being on the library path during analysis.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
mareknovotny: it is an experiment for that not working built-in types which implements org.hibernate.type.descriptor.sql.SqlTypeDescriptor. We have in EAP metadatas jars with APIs, so there is hibernate-core also for resolving the class hierarchy and to be activated that requires to have target=eap.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
mareknovotny: well, not sure. I was experimenting with parameter, but it was somehow now working in XPath expression. So I tried the real identifier string instead, I need to get it back if it works, but I suspect it won't work, but your suggestion is good. I will fix it.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: good catch.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
jsight: @mareknovotny - It should be possible to do this with parameterization. Take a peak at https://github.com/windup/windup/blob/master/rules-xml/tests/src/test/java/org/jboss/windup/rules/xml/XmlFileParameterizedTest.java

mareknovotny: cool, i didn't know that, I got it to work with windup:matches() function.

jsight: @mareknovotny - Great... I am glad to hear that the code actually works! :) Very few rules have used it so far.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: probably yes

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: probably yes

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
Ladicek: Not that I care much :-), but this can possibly affect people migrating from XXX to EAP 6, no?

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +45,109 @@
...
+        </rule>
+        <rule id="hibernate4-00003">
+            <when>
+                <xmlfile 
+                   matches="//generator/@class[windup:matches(self::node(), '{generator}')]"
+                 in="{*}.hbm.xml"/>
+            </when>
+            <perform>
+                <hint title="Identifier generator {generator} was removed in Hibernate 5" effort="1" severity="mandatory">
+                    <message>Identifier generator {generator} was removed in Hibernate 5. You can use `sequence` generator instead.</message>
```
Ladicek: Similarly to the rule `hibernate4-00002`, the `<hint title="...">` should read "Identifier generator {generator} was removed/deprecated in Hibernate 5" (yes, some of them were only deprecated) and the `<message>` should read "Use `sequence` instead."

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
mareknovotny: true, but we will need to add eap7 metadata too, so after succesfull experimenting with eap, I will change/remove it. ;)

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
mareknovotny: fixed

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: added hibernate4-xml-00005 rule

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: added hibernate4-xml-00004 rule

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +45,109 @@
...
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+            <where param="generator">
+                <matches pattern="(sequence|sequence-identity|hilo)" />
+            </where>
+        </rule>
+        <rule id="hibernate4-xml-00004">
+            <when>
+                <xmlfile matches="//property[text() = 'org.hibernate.hql.spi.PersistentTableBulkIdStrategy']" as="cfg" />
```
Ladicek: This rule is about `TemporaryTableBulkIdStrategy`, but this `<xmlfile>` mentions `PersistentTableBulkIdStrategy`. A rule for that is below.

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +45,109 @@
...
+                <filecontent pattern="hibernate.hql.bulk_id_strategy=org.hibernate.hql.spi.TemporaryTableBulkIdStrategy" filename="hibernate.properties" as="properties" />
+            </when>
+            <perform>
+                <iteration over="properties">
+                    <hint title="TemporaryTableBulkIdStrategy was replaced in Hibernate 5" effort="1" severity="mandatory">
```
Ladicek: The rule for `PersistentTableBulkIdStrategy` has a message `Class PersistentTableBulkIdStrategy was moved in Hibernate 5` and this rule has `TemporaryTableBulkIdStrategy was replaced in Hibernate 5`. The messages should IMHO be consistent.

mareknovotny: copy&paste error :(

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
Ladicek: OK, good!

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +44,113 @@
...
+                        href="https://github.com/hibernate/hibernate-orm/blob/5.0/migration-guide.adoc#deprecations"
+                        title="Removed or deprecated types in Hibernate 5" />
+                    <link
+                        href="https://access.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/7.0.beta/migration-guide/chapter-4-application-changes#hibernate_and_jpa_migration_changes"
+                        title="Hibernate and JPA migration changes" />
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+            <where param="generator">
+                <matches pattern="(sequence|sequence-identity|hilo)" />
```
Ladicek: `sequence` should be `seqhilo`. The other two are OK.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,161 @@
...
+                     <link
+                        href="https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/boot/model/relational/ExportableProducer.html#registerExportables-org.hibernate.boot.model.relational.Database-"
+                        title="Hibernate 5 ExportableProducer API" />
+                    <tag>hibernate</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hibernate4-00037">
+            <when>
+                <javaclass references="org.hibernate.envers.configuration.spi.AuditConfiguration" />
```
Ladicek: Should be `org.hibernate.envers.configuration.AuditConfiguration` (and below too)

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +44,113 @@
...
+            </where>
+        </rule>
+        <rule id="hibernate4-xml-00004">
+            <when>
+                <or>
+                    <xmlfile matches="//property[text() = 'org.hibernate.hql.spi.TemporaryTableBulkIdStrategy']" as="cfg" />
```
Ladicek: Should probably have `in="hibernate.cfg.xml"`

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -40,5 +44,113 @@
...
+                        <tag>hibernate</tag>
+                        <tag>configuration</tag>
+                    </hint>
+                </iteration>
+            </perform>
+        </rule>
+        <rule id="hibernate4-xml-00005">
+            <when>
+                <or>
+                    <xmlfile matches="//property[text() = 'org.hibernate.hql.spi.PersistentTableBulkIdStrategy']" as="cfg" />
```
Ladicek: Should probably have `in="hibernate.cfg.xml"`

Ladicek:
I ran my tests with current version of the PR (`691ada8`) and the `seqhilo` and `AuditConfiguration` are the only two remaining issues.

mareknovotny:
ok, I changed the sequence -> seqhilo and removed spi from AuditConfiguration package.

There is still the org.hibernate.type.descriptor.sql.SqlTypeDescriptor issuem which @jsight get to fixed on his local environment, but I need to re-check it. But after all we are almost done ;). Good job!

Ladicek:
Oh, right, forgot about the SqlTypeDescriptor thing, because I still don't have a test for it. I will recheck all the rules for coverage, just to be sure :-)

Ladicek:
OK, confirmed that I have tests for all the rules except of the `SqlTypeDescriptor` one. I'm writing the last remaining test now and found https://hibernate.atlassian.net/browse/HHH-10154 -- essentially, the migration issue with `SqlTypeDescriptor` applies to `JavaTypeDescriptor` as well.

Here's the commit that did the change for `SqlTypeDescriptor`: https://github.com/hibernate/hibernate-orm/commit/f6322b5cddb0063231eb590a24f9dddfbe7a1507 

Here's the commit that did the same change for `JavaTypeDescriptor`: https://github.com/hibernate/hibernate-orm/commit/e27fb3f59242f7a01cf4d0a784bc26a61c47eed2 (Looking at the commit, it looks like some builtin `JavaTypeDescriptor`s still register themselves with the registry in the constructor, e.g. `EnumJavaTypeDescriptor`, but there's no general rule. I think we should mark them all.)

(Funny thing is that the `SqlTypeDescriptor` change happened already in Hibernate 4.3.9. But EAP 6 had Hibernate 4.2 by default, so it's still relevant for migration.)

mareknovotny:
@Ladicek so if I understand correctly you suggest to mark all implementations of `SqlTypeDescriptor` and `JavaTypeDescriptor`?

From javadoc https://docs.jboss.org/hibernate/orm/5.0/javadocs/index.html?org/hibernate/type/descriptor/sql/SqlTypeDescriptor.html and https://docs.jboss.org/hibernate/orm/5.0/javadocs/index.html?org/hibernate/type/descriptor/sql/SqlTypeDescriptor.html
All Known Implemented classes are:

```
org.hibernate.spatial.dialect.mysql.MySQLGeometryTypeDescriptor
org.hibernate.spatial.dialect.h2geodb.GeoDBGeometryTypeDescriptor
org.hibernate.spatial.dialect.postgis.PGGeometryTypeDescriptor
org.hibernate.spatial.dialect.oracle.SDOGeometryTypeDescriptor
org.hibernate.spatial.dialect.sqlserver.SqlServer2008GeometryTypeDescriptor
org.hibernate.type.descriptor.converter.AttributeConverterSqlTypeDescriptorAdapter
org.hibernate.type.descriptor.sql.BigIntTypeDescriptor
org.hibernate.type.descriptor.sql.BinaryTypeDescriptor
org.hibernate.type.descriptor.sql.BitTypeDescriptor
org.hibernate.type.descriptor.sql.BlobTypeDescriptor
org.hibernate.type.descriptor.sql.BooleanTypeDescriptor
org.hibernate.type.descriptor.sql.CharTypeDescriptor
org.hibernate.type.descriptor.sql.ClobTypeDescriptor
org.hibernate.type.descriptor.sql.DateTypeDescriptor
org.hibernate.type.descriptor.sql.DecimalTypeDescriptor
org.hibernate.type.descriptor.sql.DoubleTypeDescriptor
org.hibernate.type.descriptor.sql.FloatTypeDescriptor
org.hibernate.type.descriptor.sql.IntegerTypeDescriptor
org.hibernate.type.descriptor.sql.LongNVarcharTypeDescriptor
org.hibernate.type.descriptor.sql.LongVarbinaryTypeDescriptor
org.hibernate.type.descriptor.sql.LongVarcharTypeDescriptor
org.hibernate.type.descriptor.sql.NCharTypeDescriptor
org.hibernate.type.descriptor.sql.NClobTypeDescriptor
org.hibernate.type.descriptor.sql.NumericTypeDescriptor
org.hibernate.type.descriptor.sql.NVarcharTypeDescriptor
org.hibernate.type.descriptor.sql.RealTypeDescriptor
org.hibernate.type.descriptor.sql.SmallIntTypeDescriptor
org.hibernate.type.descriptor.sql.SqlTypeDescriptorRegistry.ObjectSqlTypeDescriptor
org.hibernate.type.descriptor.sql.TimestampTypeDescriptor
org.hibernate.type.descriptor.sql.TimeTypeDescriptor
org.hibernate.type.descriptor.sql.TinyIntTypeDescriptor
org.hibernate.type.descriptor.sql.VarbinaryTypeDescriptor
org.hibernate.type.descriptor.sql.VarcharTypeDescriptor
org.hibernate.type.PostgresUUIDType.PostgresUUIDSqlTypeDescriptor


org.hibernate.spatial.JTSGeometryJavaTypeDescriptor
org.hibernate.spatial.GeolatteGeometryJavaTypeDescriptor
org.hibernate.type.descriptor.java.AbstractTypeDescriptor
org.hibernate.type.descriptor.java.BigDecimalTypeDescriptor
org.hibernate.type.descriptor.java.BigIntegerTypeDescriptor
org.hibernate.type.descriptor.java.BlobTypeDescriptor
org.hibernate.type.descriptor.java.BooleanTypeDescriptor
org.hibernate.type.descriptor.java.ByteArrayTypeDescriptor
org.hibernate.type.descriptor.java.ByteTypeDescriptor
org.hibernate.type.descriptor.java.CalendarDateTypeDescriptor
org.hibernate.type.descriptor.java.CalendarTimeTypeDescriptor
org.hibernate.type.descriptor.java.CalendarTypeDescriptor
org.hibernate.type.descriptor.java.CharacterArrayTypeDescriptor
org.hibernate.type.descriptor.java.CharacterTypeDescriptor
org.hibernate.type.descriptor.java.ClassTypeDescriptor
org.hibernate.type.descriptor.java.ClobTypeDescriptor
org.hibernate.type.descriptor.java.CurrencyTypeDescriptor
org.hibernate.type.descriptor.java.DateTypeDescriptor
org.hibernate.type.descriptor.java.DoubleTypeDescriptor
org.hibernate.type.descriptor.java.DurationJavaDescriptor
org.hibernate.type.descriptor.java.EnumJavaTypeDescriptor
org.hibernate.type.descriptor.java.FloatTypeDescriptor
org.hibernate.type.descriptor.java.InstantJavaDescriptor
org.hibernate.type.descriptor.java.IntegerTypeDescriptor
org.hibernate.type.descriptor.java.JavaTypeDescriptorRegistry.FallbackJavaTypeDescriptor
org.hibernate.type.descriptor.java.JdbcDateTypeDescriptor
org.hibernate.type.descriptor.java.JdbcTimestampTypeDescriptor
org.hibernate.type.descriptor.java.JdbcTimeTypeDescriptor
org.hibernate.type.descriptor.java.LocalDateTimeJavaDescriptor
org.hibernate.type.descriptor.java.LocalDateJavaDescriptor
org.hibernate.type.descriptor.java.LocaleTypeDescriptor
org.hibernate.type.descriptor.java.LocalTimeJavaDescriptor
org.hibernate.type.descriptor.java.LongTypeDescriptor
org.hibernate.type.descriptor.java.NClobTypeDescriptor
org.hibernate.type.descriptor.java.OffsetDateTimeJavaDescriptor
org.hibernate.type.descriptor.java.OffsetTimeJavaDescriptor
org.hibernate.type.descriptor.java.PrimitiveByteArrayTypeDescriptor
org.hibernate.type.descriptor.java.PrimitiveCharacterArrayTypeDescriptor
org.hibernate.type.descriptor.java.SerializableTypeDescriptor
org.hibernate.type.descriptor.java.ShortTypeDescriptor
org.hibernate.type.descriptor.java.StringTypeDescriptor
org.hibernate.type.descriptor.java.TimeZoneTypeDescriptor
org.hibernate.type.descriptor.java.UrlTypeDescriptor
org.hibernate.type.descriptor.java.UUIDTypeDescriptor
org.hibernate.type.descriptor.java.ZonedDateTimeJavaDescriptor
```

Ladicek:
@mareknovotny Indeed that's what I suggest :-) The solution to transitive scanning wasn't successful and we'll have to list all the built-in impls manually?

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,226 @@
...
+                    </javaclass>
+                    <javaclass references="org.hibernate.type.descriptor.java.JavaTypeDescriptor">
+                        <location>IMPLEMENTS_TYPE</location>
+                    </javaclass>
+                </or>
+                
+            </when>
+            <perform>
+                <hint title="Class org.hibernate.type.descriptor.sql.SqlTypeDescriptor change in Hibernate 5" effort="1" severity="mandatory">
+                    <message>Built-in `org.hibernate.type.descriptor.sql.SqlTypeDescriptor` or `org.hibernate.type.descriptor.java.JavaTypeDescriptor` implementations no longer auto-register themselves with `org.hibernate.type.descriptor.sql.SqlTypeDescriptorRegistry`.
```
Ladicek: It's `SqlTypeDescriptorRegistry` for `SqlTypeDescriptor` and `JavaTypeDescriptorRegistry` for `JavaTypeDescriptor`.

Ladicek: Actually, now I see that there's another rule that covers custom `*TypeDescriptor`s inheriting from the builtin ones. That one is correct.

This one is no longer useful, because if you implement the interface without inheriting from a built-in class, there's nothing you have to change. The auto-registering behavior that changed between EAP 6 and 7 was only provided by the built-in classes.

mareknovotny:
ok, I removed rule 29.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -10,9 +10,10 @@
             <addon id="org.jboss.windup.rules,windup-rules-java,2.4.0.Final" />
         </dependencies>
         <sourceTechnology id="hibernate" versionRange="[4,)" />
```
Ladicek: This should be `[4,5)` just like in `hibernate4-xml.windup`, no?

rules-reviewed/eap7/eap6/hibernate4.windup.xml
```diff
@@ -604,5 +572,198 @@
...
+                <matches
+                    pattern="(AbstractTypeDescriptor | BigDecimalTypeDescriptor | BigIntegerTypeDescriptor | BlobTypeDescriptor | BooleanTypeDescriptor | ByteArrayTypeDescriptor | ByteTypeDescriptor | CalendarDateTypeDescriptor | CalendarTimeTypeDescriptor | CalendarTypeDescriptor | CharacterArrayTypeDescriptor | CharacterTypeDescriptor | ClassTypeDescriptor | ClobTypeDescriptor | CurrencyTypeDescriptor | DateTypeDescriptor | DoubleTypeDescriptor | DurationJavaDescriptor | EnumJavaTypeDescriptor | FloatTypeDescriptor | InstantJavaDescriptor | IntegerTypeDescriptor | JavaTypeDescriptorRegistry.FallbackJavaTypeDescriptor | JdbcDateTypeDescriptor | JdbcTimestampTypeDescriptor | JdbcTimeTypeDescriptor | LocalDateTimeJavaDescriptor | LocalDateJavaDescriptor | LocaleTypeDescriptor | LocalTimeJavaDescriptor | LongTypeDescriptor | NClobTypeDescriptor | OffsetDateTimeJavaDescriptor | OffsetTimeJavaDescriptor | PrimitiveByteArrayTypeDescriptor | PrimitiveCharacterArrayTypeDescriptor | SerializableTypeDescriptor | ShortTypeDescriptor | StringTypeDescriptor | TimeZoneTypeDescriptor | UrlTypeDescriptor | UUIDTypeDescriptor | ZonedDateTimeJavaDescriptor)" />
```
Ladicek: There's only one `AbstractTypeDescriptor` in Hibernate -- in package `org.hibernate.type.descriptor.java`. So this is in wrong `<where>`.

Ladicek:
I just found that the `{Java,Sql}TypeDescriptorRegistry`-ies were added in Hibernate 4.3 (in this commit: https://github.com/hibernate/hibernate-orm/commit/ad2a9ef65159672f11091d8b6ea82752f4c2cc1c), so the migration concern with extending built-in implementations of `{Java,Sql}TypeDescriptor` is only valid for people who used Hibernate 4.3. We should probably indicate that in the migration warning.

rules-reviewed/eap7/eap6/hibernate4.windup.xml
mareknovotny: well, it doesn't matter much as we don't have that implemented :(, anyway I will add 5 there

mareknovotny:
I added also EAP 7 technology API jars to resolve class hierarchy. This needs to be enhanced later as described in https://issues.jboss.org/browse/WINDUP-957.

Ladicek:
I looked at the API jars and I found that an EAP 6 set of API jars is already in place. So here's a question.

If we added a rule that marked all classes that `implements {Java,Sql}TypeDescriptor`, it would also find all the classes that extend the built-in implementations of `{Java,Sql}TypeDescriptor` without implementing the interface directly. Right?

And we don't do that only because we don't want to mark classes that implement the interface, we only want to mark classes that extend the known built-in implementations, because only classes that extend the built-in implementations have to be changed. Right?

If so, I think that just looking for all implementations of `{Java,Sql}TypeDescriptor` and mentioning the built-in implementations only in the migration warning would be good enough. It would make the rule implementation much simpler IMHO.

If at least one answer to the above questions is "no", then I clearly misunderstood somewhere.

mareknovotny:
yes, that should cover that. The problem is that metadata is set in rules and we need target not source. So adding new api-jars is a must AFAIK. I try to simplify the rule.

Ladicek:
Hmm, that sounds strange. I'd expect that when I'm migrating _from_ technology "eap" to "eap7", the API jars for technology "eap" would be used and API jars for technology "eap7" wouldn't even be considered.

jsight:
@Ladicek It is a matter of perspective... the feature was originally implemented to assist with determining class reachability in the new app server. For that, you would want the target.

For the case that we are talking about here, the source would make more sense (just as you are saying).

I'm not really sure what the best path is for satisfying both requirements, tbh.

Ladicek:
Aha, that makes sense. I'll leave that up to you as the expert :-)

jsight:
@mareknovotny - These are ready to merge from my perspective. Feel free to merge whenever you feel like they are done. :)

mareknovotny:
ok, thanks, I will merge it now and will simplify the rule related to built-in `{Java,Sql}TypeDescriptor` conditions

PR #130, opened 2016-04-06
[WINDUPRULE-139] [WINDUPRULE-140] add rules for JBoss Logging annotations and JSF incompatibility

Ladicek:
I figured I will try to add some easier rules myself to better understand what it takes :-)


mareknovotny:
Thanks for new additions! Really appreciated.

PR #132, opened 2016-04-13
WINDUPRULE-130 adding hibernate search rules for EAP7 migration

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-130
There are still some issues - tests and probably rule definitions

### DIFF
--- a/rules-reviewed/eap7/eap6/hsearch.windup.xml
+++ b/rules-reviewed/eap7/eap6/hsearch.windup.xml
@@ -1,0 +1,1144 @@
+<?xml version="1.0"?>
+<ruleset xmlns="http://windup.jboss.org/schema/jboss-ruleset" id="hsearch" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
+            This ruleset provides analysis for migration from Hibernate Search 4.x to Hibernate Search 5.x.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.4.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.4.0.Final" />
+        </dependencies>
+        <sourceTechnology id="hibernate-search" versionRange="[4,5)" />
+        <sourceTechnology id="eap" versionRange="[6,7)" />
+        <targetTechnology id="hibernate-search5" versionRange="[5,)" />
+        <targetTechnology id="eap7" versionRange="[7,)" />
+        <tag>reviewed-2016-04-15</tag>
+    </metadata>
+    <rules>
+        <rule id="hsearch-00000">
+            <when>
+                <javaclass references="org.hibernate.search.{*}" />
+            </when>
+            <perform>
+                <classification title="Hibernate Search API usage" effort="0" severity="mandatory">
+                    <description>Hibernate Search library API used in application.</description>
+                    <tag>hibernate-search</tag>
+                </classification>
+            </perform>
+        </rule>
+        <rule id="hsearch-00001">
+            <when>
+                <javaclass references="org.hibernate.search.impl.SearchMappingBuilder" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 renamed class SearchMappingBuilder" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.impl.SearchMappingBuilder` was replaced by `org.hibernate.search.engine.spi.SearchMappingHelper` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00002">
+            <when>
+                <javaclass references="org.hibernate.search.Environment" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for Environment" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.Environment` was replaced by `org.hibernate.search.cfg.Environment` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00003">
+            <when>
+                <javaclass references="org.hibernate.search.FullTextFilter" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for FullTextFilter" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.FullTextFilter` was replaced by `org.hibernate.search.filter.FullTextFilter` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00004">
+            <when>
+                <javaclass references="org.hibernate.search.indexes.impl.DirectoryBasedIndexManager" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for DirectoryBasedIndexManager" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.indexes.impl.DirectoryBasedIndexManager` was replaced by `org.hibernate.search.indexes.spi.DirectoryBasedIndexManager` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00005">
+            <when>
+                <javaclass references="org.hibernate.search.infinispan.impl.InfinispanDirectoryProvider" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for InfinispanDirectoryProvider" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.infinispan.impl.InfinispanDirectoryProvider` was replaced by `org.hibernate.search.infinispan.spi.InfinispanDirectoryProvider` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00006">
+            <when>
+                <javaclass references="org.hibernate.search.ProjectionConstants" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for ProjectionConstants" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.ProjectionConstants` was replaced by `org.hibernate.search.engine.ProjectionConstants` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00007">
+            <when>
+                <javaclass references="org.hibernate.search.SearchException" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for SearchException" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.SearchException` was replaced by `org.hibernate.search.exception.SearchException` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00008">
+            <when>
+                <javaclass references="org.hibernate.search.spi.MassIndexerFactory" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for MassIndexerFactory" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.spi.MassIndexerFactory` was replaced by `org.hibernate.search.batchindexing.spi.MassIndexerFactory` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00009">
+            <when>
+                <javaclass references="org.hibernate.search.spi.SearchFactoryBuilder" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for SearchFactoryBuilder" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.spi.SearchFactoryBuilder` was replaced by `org.hibernate.search.spi.SearchIntegratorBuilder` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link href="http://docs.jboss.org/hibernate/search/5.5/api/org/hibernate/search/spi/SearchIntegratorBuilder.html" title="SearchIntegrationBuilder javadoc" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>        
+        <rule id="hsearch-00010">
+            <when>
+                <javaclass references="org.hibernate.search.spi.SearchFactoryIntegrator" />
+            </when>
+            <perform>
+                <hint title="Deprecated interface org.hibernate.search.spi.SearchFactoryIntegrator in Hibernate Search 5" effort="1" severity="mandatory">
+                    <message>Do not use `org.hibernate.search.spi.SearchFactoryIntegrator`. You should migrate all code to use `org.hibernate.search.spi.SearchIntegrator`.
+                    This interface is a temporary placeholder and will be removed at the next micro release!</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link href="https://docs.jboss.org/hibernate/search/5.5/api/org/hibernate/search/spi/SearchIntegrator.html"
+                        title="Hibernate Search javadoc" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>       
+        <rule id="hsearch-00011">
+            <when>
+                <javaclass references="org.hibernate.search.Version" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for Version" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.Version` was replaced by `org.hibernate.search.engine.Version` class.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <!-- Deprecated things -->
+        <rule id="hsearch-00100">
+            <when>
+                <javaclass references="org.hibernate.search.cfg.NumericFieldMapping(java.lang.String, org.hibernate.search.cfg.PropertyDescriptor, org.hibernate.search.cfg.EntityDescriptor, org.hibernate.search.cfg.SearchMapping)">
+                    <location>CONSTRUCTOR_CALL</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Constructor NumericFieldMapping.NumericFieldMapping is deprecated in Hibernate Search 5" effort="1"
+                    severity="mandatory">
+                    <message>Use `NumericFieldMapping.NumericFieldMapping(String, PropertyDescriptor, EntityDescriptor, SearchMapping)` instead.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link
+                        href="https://docs.jboss.org/hibernate/search/5.5/api/org/hibernate/search/cfg/NumericFieldMapping.html#NumericFieldMapping-org.hibernate.search.cfg.PropertyDescriptor-org.hibernate.search.cfg.EntityDescriptor-org.hibernate.search.cfg.SearchMapping-"
+                        title="Hibernate Search javadoc" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00101">
+            <when>
+                <javaclass references="org.hibernate.search.store.IndexShardingStrategy" />
+            </when>
+            <perform>
+                <hint title="Deprecated interface org.hibernate.search.store.IndexShardingStrategy in Hibernate Search 5" effort="1" severity="optional">
+                    <message>`org.hibernate.search.store.IndexShardingStrategy` interface is deprecated as of Hibernate Search 4.4. Might be removed in Hibernate Search 5.
+                    Use `org.hibernate.search.store.ShardIdentifierProvider` instead.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link
+                        href="https://docs.jboss.org/hibernate/search/5.5/api/org/hibernate/search/store/ShardIdentifierProvider.html"
+                        title="Hibernate Search javadoc" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00103">
+            <when>
+                <javaclass references="org.hibernate.search.store.Workspace" />
+            </when>
+            <perform>
+                <hint title="Deprecated interface org.hibernate.search.store.Workspace in Hibernate Search 5" effort="1" severity="optional">
+                    <message>`org.hibernate.search.store.Workspace` interface will be moved and considered as non-public API.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00104">
+            <when>
+                <javaclass references="org.hibernate.search.filter.FilterKey" />
+            </when>
+            <perform>
+                <hint title="Deprecated interface org.hibernate.search.filter.FilterKey in Hibernate Search 5" effort="1" severity="optional">
+                    <message>Custom filter keys are deprecated and are scheduled for removal in Hibernate Search 6. As of Hibernate Search 5.1, keys for caching Lucene filters are calculated automatically based on the given filter parameters.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link href="http://hibernate.org/search/documentation/migrate/5.1/#api-changes" title="Hibernate Search 5.1 API changes" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+
+        <rule id="hsearch-00106">
+            <when>
+                <javaclass references="org.hibernate.search.filter.StandardFilterKey" />
+            </when>
+            <perform>
+                <hint title="Deprecated class org.hibernate.search.filter.StandardFilterKey in Hibernate Search 5" effort="1" severity="optional">
+                    <message>Custom filter keys `StandardFilterKey` are deprecated and are scheduled for removal in Hibernate Search 6.
+                    As of Hibernate Search 5.1, keys for caching Lucene filters are calculated automatically based on the given filter parameters.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link href="http://hibernate.org/search/documentation/migrate/5.1/#api-changes" title="Hibernate Search 5.1 API changes" />
+                    <link
+                        href="https://docs.jboss.org/hibernate/search/5.5/api/index.html?org/hibernate/search/filter/StandardFilterKey.html"
+                        title="hibernate Search javadoc" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00107">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.FieldCacheType" />
+            </when>
+            <perform>
+                <hint title="Deprecated annotation FieldCacheType in Hibernate Search 5" effort="1" severity="optional">
+                    <message>Remove the `@FieldCacheType` annotation. No alternative replacement necessary as the Field Cache in Apache Lucene feature does no longer exist.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link href="https://docs.jboss.org/hibernate/search/5.5/api/org/hibernate/annotations/FieldCacheType.html"
+                        title="hibernate Search javadoc" />
+                    <tag>hibernate-search</tag>
+                    <tag>annotation</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00108">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.CacheFromIndex">
+                    <location>ANNOTATION</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Deprecated annotation CacheFromIndex in Hibernate Search 5" effort="1" severity="optional">
+                    <message>Remove the `@CacheFromIndex` annotation. No alternative replacement necessary.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <link href="https://docs.jboss.org/hibernate/search/5.5/api/org/hibernate/search/annotations/CacheFromIndex.html"
+                        title="hibernate Search javadoc for @CacheFromIndex" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>      
+        <rule id="hsearch-00109">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.Key">
+                    <location>ANNOTATION</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Deprecated annotation Key in Hibernate Search 5" effort="1" severity="optional">
+                    <message>Custom filter cache keys are a deprecated feature and are scheduled to be removed in Hibernate Search 6.
+                     As of Hibernate Search 5.1, the filter cache keys are determined automatically based on the filter parameters so it is no longer required to provide a key object. </message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <link href="http://hibernate.org/search/documentation/migrate/5.1/#api-changes" title="Hibernate Search 5.1 API changes" />
+                    <link href="https://docs.jboss.org/hibernate/search/5.5/api/index.html?org/hibernate/search/annotations/Key.html"
+                        title="Hibernate Search javadoc for @Key" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>        
+        <rule id="hsearch-00110">
+            <when>
+                <javaclass references="org.hibernate.search.backend.configuration.impl.IndexWriterSetting.{constant}" />
+            </when>
+            <perform>
+                <hint title="Deprecated enum Constants in Hibernate Search 5" effort="1" severity="optional">
+                    <message>The `IndexWriterSetting.MAX_THREAD_STATES` and `IndexWriterSetting.TERM_INDEX_INTERVAL` enum constants are deprecated.
+                    They affect which properties are read from the configuration, so the fact they they are missing means that configuration properties such as `hibernate.search.Animals.2.indexwriter.term_index_interval = default` are now ignored.
+                    The only side effect is that the property is not applied.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+            <where param="constant">
+                <matches pattern="(MAX_THREAD_STATES|TERM_INDEX_INTERVAL)" />
+            </where>
+        </rule>
+        <rule id="hsearch-00111">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.Spatial" >
+                    <location>ANNOTATION</location>
+                </javaclass>
+                <!-- https://issues.jboss.org/browse/WINDUP-974 -->
+                <!-- This doesn't find annotation with enum element specified -->
+<!--                 <javaclass references="org.hibernate.search.annotations.SpatialMode.GRID" />                 -->
+            </when>
+            <perform>
+                <hint title="Renamed SpatialMode.GRID to HASH in Hibernate Search 5" effort="1" severity="mandatory">
+                    <message>The enum value `SpatialMode.GRID` for `@Spatial` annotation was renamed to `SpatialMode.HASH`.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00112">
+            <when>
+                <javaclass references="org.hibernate.search.event.impl.FullTextIndexEventListener">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="FullTextIndexEventListener class is now final in Hibernate Search 5" effort="1" severity="mandatory">
+                    <message>The class `FullTextIndexEventListener` is now a final class. If you currently extend this class, you must find an alternate solution to achieve the same functionality. You could in some cases use an `EntityIndexingInterceptor`.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <link href="http://hibernate.org/search/documentation/migrate/5.0/#fulltextindexeventlistener-now-final" title="FullTextIndexEventListener now final" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00113">
+            <when>
+                <javaclass references="org.hibernate.search.backend.impl.jms.AbstractJMSHibernateSearchController">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Usage of internal AbstractJMSHibernateSearchController class in Hibernate Search 5" effort="1"
+                    severity="mandatory">
+                    <message>The JMS backend was depending to Hibernate ORM. This dependency was removed, so that the backend can be used in other (non ORM) environments as well.
+                    A consequence is that implementors of `org.hibernate.search.backend.impl.jms.AbstractJMSHibernateSearchController` will need to adjust to the new signature.
+                     This class is really considered internal. We suggest to take this class as an example instead of extending it.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <link href="http://hibernate.org/search/documentation/migrate/5.0/#jms-controller-api-changed" title="JMS Controller API changed" />
+
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00114">
+            <when>
+                <javaclass references="org.hibernate.search.spi.ServiceProvider">
+                    <location>IMPLEMENTS_TYPE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Implementing ServiceProvider in Hibernate Search 5" effort="1"
+                    severity="mandatory">
+                    <message>The `org.hibernate.search.spi.ServiceProvider` SPI has been refactored.
+                    If you where integrating with the old service contract, refer to the javadoc of `ServiceManager`, `Service`, `Startable` and `Stoppable` for details about the new contract.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <link
+                        href="http://docs.jboss.org/hibernate/search/5.0/api/org/hibernate/search/engine/service/spi/ServiceManager.html"
+                        title="Hibernate Search javadoc for ServiceManager" />
+                    <link href="http://docs.jboss.org/hibernate/search/5.0/api/org/hibernate/search/engine/service/spi/Service.html"
+                        title="Hibernate Search javadoc for Service" />
+                    <link
+                        href="http://docs.jboss.org/hibernate/search/5.0/api/org/hibernate/search/engine/service/spi/Stoppable.html"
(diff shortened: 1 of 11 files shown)


rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,880 @@
...
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <!-- Deprecated things -->
+        <rule id="hsearch-00100">
+            <when>
+                <javaclass references="org.hibernate.search.cfg.NumericFieldMapping(java.lang.String, org.hibernate.search.cfg.PropertyDescriptor, org.hibernate.search.cfg.EntityDescriptor, org.hibernate.search.cfg.SearchMapping)">
+                    <location>CONSTRUCTOR_CALL</location>
```
mareknovotny: not working, don't know why

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,880 @@
...
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00200">
+            <when>
+                <project>
+                    <artifact artifactId="hibernate-search-analyzers" groupId="org.hibernate" />
+                </project>
+            </when>
```
mareknovotny: project/artifact is not able to be used in ParametrizedIterationOperation like hint or classification.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,880 @@
...
+                <matches pattern="(MAX_THREAD_STATES|TERM_INDEX_INTERVAL)" />
+            </where>
+        </rule>
+        <rule id="hsearch-00111">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.Spatial" >
+                    <location>ANNOTATION</location>
+                </javaclass>
+                <!-- This doesn't find annotation with enum element specified -->
+<!--                 <javaclass references="org.hibernate.search.annotations.SpatialMode.GRID" />                 -->
```
mareknovotny: There should be possible to catch this:
`@Spatial(spatialMode = SpatialMode.GRID)`

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,880 @@
...
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00114">
+            <when>
+                <javaclass references="org.hibernate.search.engine.service.spi.ServiceProvider">
+                <!-- it doesn't catch the implements in test code -->
+                    <location>IMPLEMENTS_TYPE</location>
```
mareknovotny: not working, there is https://github.com/mareknovotny/windup-rulesets/blob/d0b82c2d4a5c3903ffcb0f8335a43f2fdd5f6cdd/rules-reviewed/eap7/eap6/tests/data/data-hsearch/CacheServiceProvider.java#L6

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,880 @@
...
+        <!-- the type of the marker must be compatible with all other values that are indexed in that same field. -->
+        <!-- For example, it was previously possible to encode a null value for numeric fields using a string "null". -->
+        <!-- This is no longer allowed. Instead, you must choose a number to represent the null value, such as -1. -->
+        <rule id="hsearch-00117">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.Field">
+                    <location>ANNOTATION</location>
+                </javaclass>
+                <javaclass references="org.hibernate.search.annotations.Field.indexNullAs">
+                    <location>METHOD_CALL</location> <!-- ??? -->
```
mareknovotny: can't express `@Field(store = Store.YES, index = Index.YES, indexNullAs=-1)`

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,880 @@
...
+                </project>
+            </when>
+            <perform>
+                <lineitem message="Maven artifact org.hibernate:hibernate-search-analyzers was removed from Hibernate Search 5. We recommend to depend on the appropriate Lucene artifact directly, for example org.apache.lucene:lucene-analyzers-common." />
+            </perform>
+        </rule>
+        <rule id="hsearch-00210">
+            <when>
+                <javaclass references="org.apache.lucene.search.SortField({*})">
+                    <location>CONSTRUCTOR_CALL</location>
```
mareknovotny: not working

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: created https://issues.jboss.org/browse/WINDUP-974

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: created https://issues.jboss.org/browse/WINDUP-974

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: I thought that this was due to the lack of an import. I added the import and it still doesn't work, though. Hmm...

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: It looks like this was an issue with the test source code.

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: Feel free to fix. :)

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: Thanks... hopefully we can get this prioritized into the current sprint.

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: Fixed here: https://github.com/windup/windup/commit/2338714a5af5f752ac5035e031872e3aad4cb16a

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: It had to do with type bindings not being found. Fixed by setting the "target" version in the rule test.

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: Can you show me the whole rule that causes the exception?

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,897 @@
...
+                <javaclass references="org.hibernate.search.backend.configuration.impl.IndexWriterSetting.{constant}" />
+            </when>
+            <perform>
+                <hint title="Deprecated Enum Constants in Hibernate Search 5" effort="1" severity="optional">
+                    <message>The `IndexWriterSetting.MAX_THREAD_STATES` and `IndexWriterSetting.TERM_INDEX_INTERVAL` enum constants are deprecated. 
+                    They affect which properties are read from the configuration, so the fact they they are missing means that configuration properties such as `hibernate.search.Animals.2.indexwriter.term_index_interval = default` are now ignored. 
+                    The only side effect is that the property is not applied.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
```
jsight: Miscellaneous is misspelled

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: confirmed, thanks!

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: great catch, it was copy&pasted from the documentation, I will fix it as now I can see it is fixed in the migration documentation too.

mareknovotny:
I still have got these failures I need to fix:
```
Failed tests: 
  WindupRulesTest.testWindupRules:140 Error with test: hsearch-00100-test
	Cause: (Rule: hsearch-00100-test) Assertion failed: hsearch-00100 hint not found!
Error with test: hsearch-00117-test
	Cause: (Rule: hsearch-00117-test) Assertion failed: hsearch-00117 hint not found!
Error with test: hsearch-00210-test
	Cause: (Rule: hsearch-00210-test) Assertion failed: hsearch-00210 hint not found!
Error with test: hsearch-00217-test
	Cause: (Rule: hsearch-00217-test) Assertion failed: hsearch-00217 hint not found!
Error with test: hsearch-00218-test
	Cause: (Rule: hsearch-00218-test) Assertion failed: hsearch-00218 hint not found!
Error with test: hsearch-00221-test
	Cause: (Rule: hsearch-00221-test) Assertion failed: hsearch-00221 hint not found!
Error with test: hsearch-00222-test
	Cause: (Rule: hsearch-00222-test) Assertion failed: hsearch-00222 hint not found!
```

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,896 @@
...
+        <description>
+            This ruleset provides analysis for migration from Hibernate Search 4.x to Hibernate Search 5.x.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.4.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.4.0.Final" />
+        </dependencies>
+        <sourceTechnology id="hibernate-search" versionRange="[4,)" />
+        <targetTechnology id="hibernate-search5" versionRange="[5,)" />
+        <targetTechnology id="eap7" versionRange="[7,)" />
```
Ladicek: Missing source technology `eap`.

jsight:
@mareknovotny - hsearch-00100-test fails because the example code uses a String as the first parameter, which is different from what the rule is looking for and the target needs to be specified as eap7. This will need a rebase to pull in the latest rulesets master, I think.

With those changes, the failures here are only:
Error with test: hsearch-00117-test
	Cause: (Rule: hsearch-00117-test) Assertion failed: hsearch-00117 hint not found!
Error with test: hsearch-00217-test
	Cause: (Rule: hsearch-00217-test) Assertion failed: hsearch-00217 hint not found!
Error with test: hsearch-00218-test
	Cause: (Rule: hsearch-00218-test) Assertion failed: hsearch-00218 hint not found!
Error with test: hsearch-00222-test
	Cause: (Rule: hsearch-00222-test) Assertion failed: hsearch-00222 hint not found!


00117 - Related to annotations. I'm working on the JIRA to add query support for this
00217 - Doesn't appear that the method call appears in the sample code?
00218 - Source doesn't have the same object type as the rule is searching for
00222 - The sample code imports something else, but it looks like it should be importing org.hibernate.search.query.dsl.QueryBuilder

I am hoping to have something to help with the annotations later today.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,896 @@
...
+        <tag>reviewed-2016-04-12</tag>
+    </metadata>
+    <rules>
+        <rule id="hsearch-00000">
+            <when>
+                <javaclass references="org.hibernate.search.{*}" />
+            </when>
+            <perform>
+                <classification title="Hibernate Search usage" effort="0" severity="mandatory">
```
Ladicek: What is this used for? I see we have `<classification>`s in `eap6[-xml].windup.xml`, but they mostly seem like they should be `<hint>`s instead. Other than that, we have no `<classification>`s anywhere.

Ladicek:
On many places, there's "The class \`XXX\` were replaced by \`YYY\`". Should be "The class _was_ replaced by".

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,896 @@
...
+                    <tag>hibernate-search</tag>
+                </classification>
+            </perform>
+        </rule>
+        <rule id="hsearch-00001">
+            <when>
+                <javaclass references="org.hibernate.search.engine.impl.SearchMappingBuilder" />
+            </when>
+            <perform>
+                <hint title="Removed org.hibernate.search.engine.impl.SearchMappingBuilder class in Hibernate Search 5" effort="1" severity="mandatory">
```
Ladicek: Similarly to other rules, this should be "Hibernate Search 5 package change for SearchMappingBuilder".

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,896 @@
...
+            <perform>
+                <classification title="Hibernate Search usage" effort="0" severity="mandatory">
+                    <description>Hibernate Search API used.</description>
+                    <tag>hibernate-search</tag>
+                </classification>
+            </perform>
+        </rule>
+        <rule id="hsearch-00001">
+            <when>
+                <javaclass references="org.hibernate.search.engine.impl.SearchMappingBuilder" />
```
Ladicek: Correct FQN here is `org.hibernate.search.impl.SearchMappingBuilder`.

This class was actually moved twice:

- from `org.hibernate.search.impl.SearchMappingBuilder` to `org.hibernate.search.engine.impl.SearchMappingBuilder` in https://github.com/hibernate/hibernate-search/commit/bdf7b30992
- from `org.hibernate.search.engine.impl.SearchMappingBuilder` to `org.hibernate.search.engine.spi.SearchMappingBuilder` in https://github.com/hibernate/hibernate-search/commit/ddd69df200

And then it was renamed from `org.hibernate.search.engine.spi.SearchMappingBuilder` to `org.hibernate.search.engine.spi.SearchMappingHelper` in https://github.com/hibernate/hibernate-search/commit/71e6b3b2a8

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: classification is different from hint as it is logical grouping of something. I used that to emphasize there is Hibernate Search API usage. I can remove that, but that is good to know still you use HSearch if nothing from other cases really needs to migrate.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,896 @@
...
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00002">
+            <when>
+                <javaclass references="org.hibernate.search.Environment" />
```
Ladicek: This class is just a bag of `public static final String`s. Those are compile-time constants, so the references will be replaced by the string values during compilation. Which means that this rule will never match when analyzing compiled applications, it only makes sense in `--sourceMode`. Not sure what to do about that, if something.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,896 @@
...
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00006">
+            <when>
+                <javaclass references="org.hibernate.search.ProjectionConstants" />
```
Ladicek: Similarly to `Environment`, this is just a bag of compile-time constants.

Ladicek:
Re the comments about compile-time constants: I'm not proposing to remove the rules. They are still useful in `--sourceMode`, even though they can't match when analyzing compiled application. (I was able to work around it in the JSF rule, because there the compile-time constant had a very specific value that I could look for without big chance of false positives. This doesn't seem like a good idea here.)

Maybe we should recommend in the Migration Guide that users should use `--sourceMode` when migrating from EAP 6 to 7 if at all possible?

Ladicek:
Regarding Lucene changes described in http://lucene.apache.org/core/4_10_2/MIGRATE.html, what do you think about incorporating rules for:

- "LUCENE-2413,LUCENE-3396: Analyzer package changes" -- analyzers are quite important, I already have that covered in my tests
- "LUCENE-2883: ValueSource changes" -- I'm not familiar with value sources, so I didn't quite bother
- "LUCENE-3283: Query parsers moved to separate module" -- I see you have that commented in the ruleset, I think these are quite important
- the Solr utils/analyzers mentioned in http://hibernate.org/search/documentation/migrate/5.0/#within-apache-lucene -- I have at least covered the `TokenizerFactory` and `TokenFilterFactory` in my tests

It's clear we can't cover _everything_ from Lucene, especially given that we're moving from Lucene 3 to Lucene 5, but we should try to cover the really important stuff.

Ladicek:
Will continue next week, still have to go through a bunch of tests I already have and check coverage.

rules-reviewed/eap7/eap6/hsearch.windup.xml
Ladicek: Absolutely, I know the difference (now; I didn't know it a couple of weeks ago, so I will probably have to file a JIRA for the existing classifications because I think that most of them, if not all, should in fact be hints). I think adding classifications like this is a _good_ idea, but the other rules we have for EAP 6 -> 7 don't do that, so I question the consistency.

mareknovotny: there could be both if that is wanted/good to see there.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: yep, after rebase to master and applying some of my commits it works too. thanks for head up.

mareknovotny:
I sorted all comments except additional Lucene rule coverage from https://github.com/windup/windup-rulesets/pull/132#issuecomment-210481593

Now there are 2 failing tests for waiting on resolving https://issues.jboss.org/browse/WINDUP-974 :
- hsearch-00117-test
- hsearch-00116-test

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,899 @@
...
+    <metadata>
+        <description>
+            This ruleset provides analysis for migration from Hibernate Search 4.x to Hibernate Search 5.x.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.4.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.4.0.Final" />
+        </dependencies>
+        <sourceTechnology id="hibernate-search" versionRange="[4,)" />
+        <sourceTechnology id="eap" versionRange="[6,)" />
```
Ladicek: Both `<sourceTechnology>`-ies should have an upper limit on the version. I know, I know, the ranges don't really mean anything, but if they're there, they should be correct.

```xml
<sourceTechnology id="hibernate-search" versionRange="[4,5)" />
<sourceTechnology id="eap" versionRange="[6,7)" />
```

mareknovotny: no big deal ;), you are right.

mareknovotny:
@Ladicek wrt https://github.com/windup/windup-rulesets/pull/132#issuecomment-210481593

>     "LUCENE-2413,LUCENE-3396: Analyzer package changes" -- analyzers are quite important, I already have that covered in my tests

I will add them

>    "LUCENE-2883: ValueSource changes" -- I'm not familiar with value sources, so I didn't quite bother

I will add it

>    "LUCENE-3283: Query parsers moved to separate module" -- I see you have that commented in the ruleset, I think these are quite important

yes, they are all there as hsearch-00224, hsearch-00225 and hsearch-00226

>   the Solr utils/analyzers mentioned in http://hibernate.org/search/documentation/migrate/5.0/#within-apache-lucene -- I have at least covered the TokenizerFactory and TokenFilterFactory in my tests

I will add them

Ladicek:
I found that there is a couple of duplicate rules:

- `hsearch-00009` and `hsearch-00105`
- `hsearch-00010` and `hsearch-00102`
- `hsearch-00108` and `hsearch-00119`
- `hsearch-00109` and `hsearch-00118`
- `hsearch-00200` and `hsearch-00201` -- but I'm really not sure here, I've never encountered these `<when>` clauses before

Also not sure why rules `hsearch-00225` and `hsearch-00226` are not integrated into rule `hsearch-00224` -- it doesn't seem like that would make any difference.

mareknovotny:
good catch for duplicated rules:

>  hsearch-00009 and hsearch-00105
> hsearch-00010 and hsearch-00102
> hsearch-00108 and hsearch-00119
> hsearch-00109 and hsearch-00118 

hsearch-00200 and hsearch-00201 are different rules as first adds for project the lineitem message and second adds the hint into pom.xml. Unfortunately it can't be done in one rule as I originally had. That needs a change in Windup code so now it is a workaround for this case.

 hsearch-00224, hsearch-00225 and hsearch-00226 are different case, they recommend different things in message elements, so I can't create dynamic suggestion based on found class in condition.

Ladicek:
Ah, now I see the difference in `hsearch-00225` and `hsearch-00226` -- those classes were not just moved into a different package, but also renamed in the process (`QueryParserToken` to `Token` and `QueryParserTokenMgrError` to `TokenMgrError`). Right, these can't be merged into `hsearch-00224`, sorry about the noise.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,899 @@
...
+            </when>
+            <perform>
+                <hint title="Indexing of id Fields of Embedded Relations" effort="1" severity="mandatory">
+                    <message><![CDATA[
+                     When using an `@IndexedEmbedded` annotation to include fields from a related entity,
+                     the id of the related entity is no longer included.
+
+                     You can enable the inclusion of the id by using the `includeEmbeddedObjectId` attribute of the `@IndexedEmbedded` annotation.
+
+                    ```java
```
Ladicek: This seems to be not supported in Windup. In the migration report, I see a code block beginning with "\`\`\`java" and a huge indentation.

Ladicek:
I think there's no rule for http://hibernate.org/search/documentation/migrate/5.3/#the-faceting-api-and-engine -- granted, it's not obvious how to construct such rule, because the change is "you previously didn't have to add any kind of annotation when you used faceting and now you have to", but there's one option: detect if the user is using faceting and if she does, provide a warning that `@Facet(s)` annotations are now needed. To detect when faceting is used, one can look for:

- `org.hibernate.search.query.dsl.QueryBuilder.facet()`
- `org.hibernate.search.FullTextQuery.getFacetManager()`
- `org.hibernate.search.query.facet.{class}`, where `{class}` is one of `Facet`, `FacetingRequest`, `FacetSelection`, `FacetSortOrder`, `RangeFacet`

Ladicek:
There's a bunch of `Deprecated Interface` hints. Shouldn't be lower-case "interface", `Deprecated interface`?

Ladicek:
Rule `hsearch-00109` could/should link to http://hibernate.org/search/documentation/migrate/5.1/#api-changes (possibly also rules `hsearch-00104` and `hsearch-00106`, since they seem to be related).

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,899 @@
...
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscellaneous Hibernate Search Changes" />
+                </hint>
+
+            </perform>
+        </rule>
+        <rule id="hsearch-00211">
+            <when>
+                <javaclass references="org.hibernate.search.cfg.ContainedInMapping.numericField{*}">
```
Ladicek: This won't catch a fluent-style usage, e.g.

```java
searchMapping.entity(Person.class)
        .indexed()
        .property("name", ElementType.FIELD)
        .containedIn()
        .numericField();
```

Here, the `containedIn()` call returns `ContainedInMapping`, but Windup doesn't know that, so doesn't apply the rule.

I don't really know what we can do in this case. Can we even do something?

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,899 @@
...
+                <javaclass references="org.hibernate.search.event.impl.FullTextIndexEventListener">
+                    <location>INHERITANCE</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="FullTextIndexEventListener class is now final in Hibernate Search 5" effort="1" severity="mandatory">
+                    <message>The class `FullTextIndexEventListener` is now a final class. If you currently extend this class, you must find an alternate solution to achieve the same functionality. You could in some cases use an `EntityIndexingInterceptor`.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
```
Ladicek: Could also link to http://hibernate.org/search/documentation/migrate/5.0/#fulltextindexeventlistener-now-final ?

rules-reviewed/eap7/eap6/hsearch.windup.xml
jsight: Windup will catch cases like that if the method chain is properly imported and the appropriate types are in the analysis classpath.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,899 @@
...
+            </when>
+            <perform>
+                <hint title="Usage of internal AbstractJMSHibernateSearchController class in Hibernate Search 5" effort="1"
+                    severity="mandatory">
+                    <message>The JMS backend was depending to Hibernate ORM. This dependency was removed, so that the backend can be used in other (non ORM) environments as well.
+                    A consequence is that implementors of `org.hibernate.search.backend.impl.jms.AbstractJMSHibernateSearchController` will need to adjust to the new signature.
+                     This class is really considered internal. We suggest to take this class as an example instead of extending it.</message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscelanous Changes in Hibernate Search 5.x" />
```
Ladicek: Could also link to http://hibernate.org/search/documentation/migrate/5.0/#jms-controller-api-changed ?

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,899 @@
...
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00114">
+            <when>
+                <javaclass references="org.hibernate.search.engine.service.spi.ServiceProvider">
```
Ladicek: Correct FQN is `org.hibernate.search.spi.ServiceProvider`. It was removed in commit https://github.com/hibernate/hibernate-search/commit/6524d5233928c0ca691e79cc0f28a6b6116d3b1b

rules-reviewed/eap7/eap6/hsearch.windup.xml
Ladicek: > and the appropriate types are in the analysis classpath

Which they typically aren't, right?

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: @jsight example is if you combine hsearch-00200 and hsearch-00201 into one rule (perform operations)

mareknovotny:
I added rule `hsearch-000118` https://github.com/windup/windup-rulesets/pull/132/commits/1475c98325d85c01c6a6e00b95086acc6a768b8b#diff-7e120243d68acaa9a3c21171decdbeb6R530 for Faceting wrt https://github.com/windup/windup-rulesets/pull/132#issuecomment-211914301

Fixed *Interface* to *interface*.

I added all additional links as suggested.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1149 @@
...
+        </rule>
+        <rule id="hsearch-00118">
+            <when>
+                <or>
+                    <javaclass references="org.hibernate.search.query.dsl.QueryBuilder.facet{*}">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="org.hibernate.search.query.dsl.QueryBuilder.facet{*}">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
```
Ladicek: Looks duplicate :-)

mareknovotny: oops, last minute copy&paste error :)

mareknovotny:
I tried to use https://github.com/windup/windup/pull/874 to get annotation filtering into rules, but still fails

```
Failed tests: 
  WindupRulesTest.testWindupRules:143 Error with test: hsearch-00116-test
	Cause: (Rule: hsearch-00116-test) Assertion failed: hsearch-00116 hint not found!
Error with test: hsearch-00117-test
	Cause: (Rule: hsearch-00117-test) Assertion failed: hsearch-00117 hint not found!
```

jsight:
I will look at this tomorrow (or earlier).

jsight:
The second one is a bug in the hint test. The () characters needs to be escaped.

The first one is the rule itself. It is really a more complex rule than what you are implying as it is conditional on two annotations existing on the field. I need to play with that one a little more.

jsight:
@mareknovotny - I have updated the windup core PR with changes to support this, and I have sent the following PR to your repo:

https://github.com/mareknovotny/windup-rulesets/pull/2

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00009">
+            <when>
+                <javaclass references="org.hibernate.search.spi.SearchFactoryBuilder" />
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 package change for SearchFactoryBuilder" effort="1" severity="mandatory">
+                    <message>The class `org.hibernate.search.spi.SearchFactoryBuilder` was replaced by `org.hibernate.search.spi.SearchIntegratorBuilder` class.</message>
```
Ladicek: In that case, it's not just a "package change", no?

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+        <!-- o.a.l.queryParser.QueryParser -> o.a.l.queryparser.classic.QueryParser -->
+        <!-- o.a.l.queryParser.QueryParserBase -> o.a.l.queryparser.classic.QueryParserBase -->
+        <!-- o.a.l.queryParser.QueryParserConstants -> o.a.l.queryparser.classic.QueryParserConstants -->
+        <!-- o.a.l.queryParser.QueryParserTokenManager -> o.a.l.queryparser.classic.QueryParserTokenManager -->
+        <rule id="hsearch-00224">
+            <when>
+                <javaclass references="org.apache.lucene.queryParser.{type}" />
+            </when>
+            <perform>
+                <hint title="Renamed Lucene class {type} in Lucene 4.x" effort="1" severity="mandatory">
```
Ladicek: This is just a package change, no?

mareknovotny: yes, thanks for catching this

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+        </rule>
+<!--     o.a.l.analysis.TeeSinkTokenFilter -> o.a.l.analysis.sinks.TeeSinkTokenFilter -->
+        <rule id="hsearch-00230">
+            <when>
+                <javaclass references="org.apache.lucene.analysis.TeeSinkTokenFilter" />
+            </when>
+            <perform>
+                <hint title="Renamed package for TeeSinkTokenFilter in Lucene 4.x" effort="1" severity="mandatory">
+                    <message>Lucene's core and contrib analyzers, along with Solr's analyzers, were consolidated into lucene/analysis module.
+
```
Ladicek: `WordListLoader` is actually called `WordlistLoader`

Ladicek: `TokenizerFactory` is good, `TokenizerFilterFactory` is bad. It's `TokenFilterFactory` -- it's also written in the comment above the rule.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Changes in indexing null values" effort="1" severity="mandatory">
+                    <message>When using `@Field(indexNullAs=)` to encode a null marker value in the index, the type of the marker must be compatible with all other values that are indexed in that same field.
+                    For example, it was previously possible to encode a null value for numeric fields using a string _null_.
+                    This is no longer allowed. Instead, you must choose a number to represent the null value, such as -1. </message>
+                    <link
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_miscellaneous_hibernate_search_changes"
+                        title="Miscellaneous Changes in Hibernate Search 5.x" />
```
Ladicek: Could also link to http://hibernate.org/search/documentation/migrate/5.5/#null-value-tokens-for-numeric-fields

Ladicek:
Unless I'm highly mistaken, there's no rule to cover http://hibernate.org/search/documentation/migrate/5.0/#number-and-date-index-format It's quite important, the page even says "The change of encoding for number and dates is probably the most important user facing change".

Does the XML rule language allow expressing a condition like "a field of type `Date`, `Calendar`, `int`, `Integer`, `long`, `Long`, `float`, `Float`, `double` or `Double` annotated by `@Field`"?

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+                    </javaclass>
+                    <javaclass references="org.hibernate.search.FullTextQuery.getFacetManager{*}">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="org.hibernate.search.query.facet.{class}" />
+                </or>
+            </when>
+            <perform>
+                <hint title="Improvements in Faceting Engine" effort="1" severity="mandatory">
+                    <message>Significant improvements were made to the faceting engine. Most of the changes do not affect the API. The one notable exception is that you must now annotate any fields you intend to use for faceting with the @Facet or @Facets annotation.</message>
```
Ladicek: The `@Facet` and `@Facets` annotations could be highlighted as code :-)

mareknovotny:
I think wrappers and others are easy, not sure about primitive types. I will try it

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -825,7 +825,7 @@ fulltextQuery.setSort(new Sort(new SortField("title", SortField.Type.STRING)))
                 <javaclass references="org.apache.lucene.queryParser.{type}" />
             </when>
             <perform>
-                <hint title="Renamed Lucene class {type} in Lucene 4.x" effort="1" severity="mandatory">
+                <hint title="Package name changed for Lucene class {type} in Lucene 4.x" effort="1" severity="mandatory">
```
Ladicek: The other rules for package changes in Lucene say "Renamed package for KeywordAnalyzer in Lucene 4.x". I think this one could also read "Renamed package for {type} in Lucene 4.x".

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+        <rule id="hsearch-00211">
+            <when>
+                <javaclass references="org.hibernate.search.cfg.ContainedInMapping.numericField{*}">
+                    <location>METHOD_CALL</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Deprecated ContainedInMapping#numericField() method in Hibernate Search 5" effort="1" severity="mandatory">
+                    <message>Method `ContainedInMapping#numericField()` has been deprecated and is scheduled for removal. Invoke `ContainedInMapping#field().numericField()` instead.
+                     instead.</message>
```
Ladicek: Too many "instead"s at the end :-)

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+        <!-- org.hibernate.search.cfg.IndexedMapping.cacheFromIndex(FieldCacheType...) -->
+        <rule id="hsearch-00214">
+            <when>
+                <javaclass references="org.hibernate.search.cfg.IndexedMapping.cacheFromIndex{*}">
+                    <location>METHOD_CALL</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Deprecated method in Hibernate Search 5" effort="1" severity="mandatory">
+                    <message>This will be removed with no replacement.</message>
```
Ladicek: Other rules have more specific hint titles/messages.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00107">
+            <when>
+                <javaclass references="org.hibernate.search.annotations.FieldCacheType" />
+            </when>
+            <perform>
+                <hint title="Deprecated annotation FieldCacheType in Hibernate Search 5" effort="1" severity="optional">
+                    <message>Remove the `@FieldCacheType` annotation. No alternative replacement necessary as the Field Cache in Apache Lucene feature does no longer exist.</message>
```
Ladicek: `FieldCacheType` is not an annotation, it's an enum that is used as an attribute of the `@CacheFromIndex` annotation.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00110">
+            <when>
+                <javaclass references="org.hibernate.search.backend.configuration.impl.IndexWriterSetting.{constant}" />
+            </when>
+            <perform>
+                <hint title="Deprecated enum Constants in Hibernate Search 5" effort="1" severity="optional">
```
Ladicek: Could mention `IndexWriterSetting`. Also "Constants" should probably be lower-case.

rules-reviewed/eap7/eap6/hsearch.windup.xml
Ladicek: I'd like to understand why is this even here. First, the Hibernate Search version shipped with EAP 6 doesn't seem to have this constructor, there's only `NumericFieldMapping(PropertyDescriptor, EntityDescriptor, SearchMapping)`. Second, the message basically says "instead of `NumericFieldMapping(String, PropertyDescriptor, EntityDescriptor, SearchMapping)`, use `NumericFieldMapping(String, PropertyDescriptor, EntityDescriptor, SearchMapping)`", which doesn't make a lot of sense.

Ladicek:
Aha, I just spotted `hsearch-00116`, looks like using `@NumericField` was previously mandatory. Makes sense.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: my bad, it should be read as use `NumericFieldMapping(String, PropertyDescriptor, EntityDescriptor, SearchMapping)` instead of `NumericFieldMapping(PropertyDescriptor, EntityDescriptor, SearchMapping)`

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+        </rule>
+        <!-- org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity.getFieldCacheOption() -->
+        <rule id="hsearch-00217">
+            <when>
+                <javaclass references="org.hibernate.search.engine.spi.DocumentBuilderIndexedEntity.getFieldCacheOption{*}">
+                    <location>METHOD_CALL</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Deprecated method in Hibernate Search 5" effort="1" severity="mandatory">
```
Ladicek: Other rules have more specific hint titles.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1153 @@
...
+        <!-- org.hibernate.search.spi.BuildContext.getIndexingStrategy() -->
+        <!-- Scheduled for removal. Use BuildContext.getIndexingMode() instead. -->
+        <rule id="hsearch-00218">
+            <when>
+                <javaclass references="org.hibernate.search.spi.BuildContext.getIndexingStrategy({*}">
+                    <location>METHOD_CALL</location>
+                </javaclass>
+            </when>
+            <perform>
+                <hint title="Deprecated method in Hibernate Search 5" effort="1" severity="mandatory">
```
Ladicek: Other rules have more specific hint titles.

Ladicek:
OK, so I have tests for all the main rules (those describing changes from all the migration guides) already done and today, I've almost finished tests for the rules for deprecated elements. Hopefully I'll get all the tests done by tomorrow.

mareknovotny:
I fixed additionally links to docs and hsearch-00111 to use annotation-literal instead of broader annotation only.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1197 @@
...
+                    <link href="http://hibernate.org/search/documentation/migrate/5.1/#api-changes" title="Hibernate Search 5.1 API changes" />
+                    <link href="https://docs.jboss.org/hibernate/search/5.5/api/index.html?org/hibernate/search/annotations/Key.html"
+                        title="Hibernate Search javadoc for @Key" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="hsearch-00110">
+            <when>
+                <javaclass references="org.hibernate.search.backend.configuration.impl.IndexWriterSetting.{constant}" />
```
Ladicek: This doesn't seem to fire for me. The test only uses an `import` of this (which is wrong BTW and couldn't even compile, it only makes sense as `import static`), which isn't very realistic. My test has this:

```java
import org.hibernate.search.backend.configuration.impl.IndexWriterSetting;

...
        System.out.println(IndexWriterSetting.TERM_INDEX_INTERVAL);
...
```

That's not very realistic either :-), but should be a good model of how Java usage of this could look like. It also seems that these `IndexWriterSetting`-s are often used as keys to a map of settings (or possibly even in Hibernate configuration XML/.properties).

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1197 @@
...
+                        title="Advanced Integrator Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <!-- org.hibernate.search.spi.BuildContext.getIndexingStrategy() -->
+        <!-- Scheduled for removal. Use BuildContext.getIndexingMode() instead. -->
+        <rule id="hsearch-00218">
+            <when>
+                <javaclass references="org.hibernate.search.spi.BuildContext.getIndexingStrategy({*}">
```
Ladicek: Why does this `getIndexingStrategy({*}` have the opening paren? Other rules for deprecated methods don't have that.

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1197 @@
...
+                        title="Miscellaneous Hibernate Search Changes" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <!-- org.hibernate.search.MassIndexer.threadsForSubsequentFetching(int) -->
+        <!-- Being ignored: this method will be removed. -->
+        <rule id="hsearch-00221">
+            <when>
+                <javaclass references="org.hibernate.search.MassIndexer.threadsForSubsequentFetching{*}">
```
Ladicek: There's also a deprecated method `threadsForIndexWriter` in EAP 6's Hibernate Search. I don't see a rule for that, presumably because that method was already removed in EAP 7's Hibernate Search. WDYT about adding a rule? (I don't even want to guess how many similar cases we miss...)

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1197 @@
...
+                        href="https://access.qa.redhat.com/documentation/en/red-hat-jboss-enterprise-application-platform/version-7.0/migration-guide/#migrate_hibernate_search_changes"
+                        title="Changes in Hibernate Search 5.x" />
+                    <tag>hibernate-search</tag>
+                </hint>
+            </perform>
+        </rule>
+        <!-- Deprecated things -->
+        <rule id="hsearch-00100">
+            <when>
+                <javaclass references="org.hibernate.search.cfg.NumericFieldMapping(org.hibernate.search.cfg.PropertyDescriptor, org.hibernate.search.cfg.EntityDescriptor, org.hibernate.search.cfg.SearchMapping)">
```
Ladicek: I wonder how much is this related to rule `hsearch-00211`. I mean, I don't see a problem in keeping this rule, but it seems not very useful.

mareknovotny: not sure, a similarity is there, but it  was listed as different case I think.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: that could be checked with some tool like https://java.net/projects/jascc or http://docs.oracle.com/javame/test-tools/sigtest/2_2/html/check_using.html. I didn't do that. 
I will add `threadsForIndexWriter` rule for sure.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: there is no reason AFAIK.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: right, I surely meant to use `import static` in the test code, so my bad as tests are sometimes not checked by java compilator due missing dependencies.

I will try to fix the usage anyway.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: we need to document this in migration doc too, @bergerhoffer and @sgilda heads up ;)

rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -0,0 +1,1197 @@
...
+                    </javaclass>
+                    <javaclass references="org.hibernate.search.annotations.Field">
+                        <location>ANNOTATION</location>
+                        <annotation-literal name="index" pattern="Index.YES"/>
+                        <annotation-type pattern="org.hibernate.search.annotations.NumericField"/>
+                    </javaclass>
+                </or>
+            </when>
+            <perform>
+                <hint title="Hibernate Search 5 changes indexing null values" effort="1" severity="mandatory">
```
Ladicek: This is actually about changes in indexing _numeric fields_, not null values.

Ladicek: Also this rule doesn't catch `Date` and `Calendar` fields annotated with `@DateBridge`/`@CalendarBridge`.

Ladicek:
My current status: I've implemented tests for all rules that are currently present, except of:

- `hsearch-00200` -- not sure how to test, not even sure what it means (I added a test for `hsearch-00201` which is about the `pom.xml`)
- `hsearch-00216`, `hsearch-00219`, `hsearch-00224` -- `QueryParserBase`, `hsearch-00225`, `hsearch-00226`, `hsearch-00236` -- these classes were not shipped with EAP 6's Hibernate Search / Lucene

There's a couple of unresolved things, too:

- The rules for Lucene's `QueryParser` changes uses a different hint message than the other Lucene rules.
- We still don't cover numeric/Date/Calendar fiels annotated by `@Field` (but not annotated by `@NumericField`/`@DateBridge`/`@CalendarBridge`), that's fairly important
- The hint of rule `hsearch-00116` is wrong. The rule is also incomplete in that it doesn't catch `Date`/`Calendar` fields annotated with `@DateBridge`/`@CalendarBridge`, even though the text of the hint speaks about that.
- Rules `hsearch-00100` and `hsearch-00110` don't fire for me, using a naive `System.out.println(...)` approach.
- Rule `hsearch-00218` has unnecessary opening paren in `getIndexingStrategy({*}`.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: it does, look line below ;)

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: OK, I added both annotations there.

mareknovotny:
> There's a couple of unresolved things, too:
> 
>     The rules for Lucene's QueryParser changes uses a different hint message than the other Lucene rules.
What rule id is that?

>     We still don't cover numeric/Date/Calendar fields annotated by @Field (but not annotated by @NumericField/@DateBridge/@CalendarBridge), that's fairly important
I haven't added this as the condition is hard to create for every possible cases and to express primitive type is hard to. I would keep this aside from others issues you drew here .

>     The hint of rule hsearch-00116 is wrong. The rule is also incomplete in that it doesn't catch Date/Calendar fields annotated with @DateBridge/@CalendarBridge, even though the text of the hint speaks about that.
What's wrong? I added Date/Calendar to it.

>     Rules hsearch-00100 and hsearch-00110 don't fire for me, using a naive System.out.println(...) approach.
What's is your testcase for hsearch-00100?

>     Rule hsearch-00218 has unnecessary opening paren in getIndexingStrategy({*}.
fixed.

Ladicek:
>> The rules for Lucene's QueryParser changes uses a different hint message than the other Lucene rules.
>
> What rule id is that?

Rule `hsearch-00224`.

>> We still don't cover numeric/Date/Calendar fields annotated by @Field (but not annotated by @NumericField/@DateBridge/@CalendarBridge), that's fairly important
>
> I haven't added this as the condition is hard to create for every possible cases and to express primitive type is hard to. I would keep this aside from others issues you drew here .

I see that you changed rule `hsearch-00119` slightly. I don't fully understand how it is supposed to work, but it matches for me on a seemingly innocent class

```java
@Entity
@Indexed
public class Author {
    @Id
    @GeneratedValue
    private Integer id;

    @Field
    private String firstName;

    @Field
    private String lastName;

    @ContainedIn
    @OneToMany(mappedBy = "author")
    private Set<Book> books;
}
```

The match is probably caused by this part of the condition:

```xml
                    <and>
                        <javaclass references="org.hibernate.search.annotations.Field">
                            <location>ANNOTATION</location>
                        </javaclass>
                        <javaclass references="java.lang.{wrapper}">
                            <location>FIELD_DECLARATION</location>
                        </javaclass>
                    </and>
```

Indeed I _do_ have a "field declaration" of type `java.lang.Integer`, and indeed I also _do_ have an element annotated with `@Field`. Who cares that those two are not really related at all :-)

This gave me an idea, though: we could merge rules `hsearch-00116` and `hsearch-00119` together and just report "Changes in indexing numeric and date values in Hibernate Search 5" on _all_ fields annotated with `@Field` or `@Fields`. WDYT? It's an overkill, but if the XML rule language doesn't give us proper tools, we'll have to accept suboptimal solutions.

>> The hint of rule hsearch-00116 is wrong. The rule is also incomplete in that it doesn't catch Date/Calendar fields annotated with @DateBridge/@CalendarBridge, even though the text of the hint speaks about that.
>
> What's wrong? I added Date/Calendar to it.

The rule `hsearch-00116` seems good now.

>> Rules hsearch-00100 and hsearch-00110 don't fire for me, using a naive System.out.println(...) approach.
>
> What's your testcase for hsearch-00100?

For both rules, I have a simple `System.out.println` call:

```java
        System.out.println(new NumericFieldMapping(new PropertyDescriptor(null, null), new EntityDescriptor(), new SearchMapping()));
        System.out.println(IndexWriterSetting.TERM_INDEX_INTERVAL);
```

None of them involve compile-time constants, and indeed in the decompiled Java source, I can see both statements. Not sure why they aren't caught.

>> Rule hsearch-00218 has unnecessary opening paren in getIndexingStrategy({*}.
>
> fixed

OK, thanks.

jsight:
For hsearch-00119, the annotation filter should actually be a part of the javaclass condition for field declaration, since the two are intended to be related.

Ladicek:
@jsight Right, makes sense.

BTW, here's the condition we would probably need for identifying fields affected by http://hibernate.org/search/documentation/migrate/5.0/#number-and-date-index-format:

- a field declaration
- that is of given type (`int`, `long`, `float`, `double`, their wrapper types, `Date` or `Calendar`)
- and is annotated using `@Field` or `@Fields`
- and _isn't_ annotated using `@NumericField` nor `@NumericFields` nor `@DateBridge` nor `@CalendarBridge` nor `@Id`

(And I think there are other ways of configuring indexing, not just annotations. Oh well.)

mareknovotny:
I tried @jsight's suggestion and for me it made sense too, but I am doing something wrong or it doesn't work:
```xml
          <when>
               <or>
                 <javaclass references="java.util.{date}">
                        <annotation-type pattern="org.hibernate.search.annotations.Field" />
                    </javaclass>
                    <javaclass references="java.lang.{wrapper}">
                        <annotation-type pattern="org.hibernate.search.annotations.Field" />
                    </javaclass>
                </or>
            </when>
`
let aside that the condition should be more selective for now.

mareknovotny:
ok, I took another approach suggested by lthon - I merged hsearch-00116 with 00119 this will probably reduces the need for negating conditions for hsearch-00119 where hsearch-00116 takes action.

So now test for hsearch-00116 passes, but when it works correctly there should be 2 more occurrences than only 2.

Secondly we agreed on our regular mtg that we could merge this PR with knowing there are 2 issues - hsearch-00110 and hsearch-00116 - I will create jiras for them. WDYT @Ladicek ?

Ladicek:
I think this PR is mostly done and requires just a couple of finishing touches. If you feel like merging it now and adding those finishing touches later, I have no objections.

mareknovotny:
I created https://issues.jboss.org/browse/WINDUPRULE-152 and https://issues.jboss.org/browse/WINDUPRULE-151

PR #133, opened 2016-04-21
WINDUPRULE-137  Catchall rules should have Optional severity

OndraZizka:
I'm giving them "Potential" severity, as it seems that's what the user really wants. Correct me if I am wrong.

### DIFF
--- a/rules-reviewed/eap6/uncategorized/generic-catchall.windup.xml
+++ b/rules-reviewed/eap6/uncategorized/generic-catchall.windup.xml
@@ -30,10 +30,10 @@
                         </not>
                     </when>
                     <perform>
-                        <classification title="Doug Lea Concurrency util" severity="mandatory">
+                        <classification title="Doug Lea Concurrency util" severity="potential">
                             <tag>catchall</tag>
                         </classification>
-                        <hint effort="0" severity="mandatory" title="Doug Lea Concurrency type reference">
+                        <hint effort="0" severity="potential" title="Doug Lea Concurrency type reference">
                             <message>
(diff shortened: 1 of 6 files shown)


rules-reviewed/eap6/uncategorized/generic-catchall.windup.xml
```diff
@@ -30,10 +30,10 @@
                         </not>
                     </when>
                     <perform>
-                        <classification title="Doug Lea Concurrency util" severity="mandatory">
+                        <classification title="Doug Lea Concurrency util" severity="potential">
```
jsight: IMO, this one should just be optional. I can't imagine a reason for it to be mandatory for an EAP migration.

jsight:
I agree. Can we mention you at the meeting to make sure we are all on the same page?

jsight:
This just needs a rebase, I think.

rules-reviewed/eap6/uncategorized/generic-catchall.windup.xml
OndraZizka: Optional and not potential? Potential is more than optional? And does "potential" imply that it's a mandatory change? I'm a bit confused.

OndraZizka:
Rebased.

rules-reviewed/eap6/uncategorized/generic-catchall.windup.xml
jsight: Well, it is really weird for this to be a catchall rule to begin with, IMO.

mareknovotny: Frankly I am not against to recategorized it. Just for a reminding, these all catch rules was to mimic the Windup 0.7 feature which identified some general packages.

PR #137, opened 2016-04-25
WINDUPRULE-150 Create new rule for empty JSP imports

OndraZizka:


rules-reviewed/eap6/jsp/jsp.windup.xml
```diff
@@ -0,0 +1,31 @@
...
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.3.0.Final" />
+        </dependencies>
+        <targetTechnology id="eap" versionRange="[6,)" />
+        <targetTechnology id="java-ee" versionRange="[6,)" />
+        <tag>reviewed-2016-04-25</tag>
+    </metadata>
+    <rules>
+        <rule id="jsp-01000">
+            <when>
+                <filecontent filename="{*}.jsp" pattern="import=&quot;&quot;" />
```
jsight: Shouldn't there be a version of this for singlequote imports as well?

OndraZizka: I have never seen single quotes in <@page import,  is that even legal?

jsight: It works in wildfly and I don't know any reason why it wouldn't be legal. Single-quotes and double-quotes are typically interchangeable in contexts like this.

PR #138, opened 2016-04-28
Fix windup.version

OndraZizka:


pom.xml
```diff
@@ -22,7 +22,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>1.7</maven.compiler.source>
         <maven.compiler.target>1.7</maven.compiler.target>
-        <version.windup>2.5.0.CR3</version.windup>
+        <version.windup>${project.version}</version.windup>
```
mareknovotny: it is better to use 2.5.0-SNAPSHOT instead of project.version as that could be evaluated differently than you expects in some cases in multi module build where parent has got different value set see http://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Interpolation

mareknovotny:
I did the 2.5.0-SNAPSHOT value update

PR #140, opened 2016-04-29
Remove weblogic.security.ignore

jsight:


mareknovotny:
it seems there is a contradiction for the rule if it is in generated code or used in real code as a reference.

PR #143, opened 2016-05-03
Winduprule-145

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-145


Ladicek:
Looks good to me, but I only did a cursory check.

PR #150, opened 2016-05-26
    WINDUPRULE-164  Clean up and merge duplicate rules

OndraZizka:


rules-reviewed/eap6/jboss-eap5/jboss-eap5-xml.windup.xml
```diff
@@ -139,12 +126,28 @@
...
+                        A JBoss specific EAR descriptor allows extensions to Java EE EAR archives configuration.
+                    </description>
+                    <link title="JBoss DTD's" href="http://www.jboss.org/j2ee/dtd/"/>
+                    <tag>jboss-eap5</tag>
+                </classification>
+            </perform>
+        </rule>
         <rule id="jboss-eap5-xml-08000">
             <when>
                 <xmlfile matches="//*[local-name()='jboss-app']" in="jboss-app.xml"/>
```
jsight: A not condition based upon the other rule could be used here to prevent it from appearing twice.

rules-reviewed/eap6/websphere/websphere-xml.windup.xml
```diff
@@ -145,28 +145,33 @@
...
                         </message>
-                        <link title="Configure the EAP6 Web Subsystem" href="https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6.4/html-single/Administration_and_Configuration_Guide/index.html#chap-Web_Subsystem"/>
+                        <link title="Configure the EAP 6 Web Subsystem" href="https://access.redhat.com/documentation/en-US/JBoss_Enterprise_Application_Platform/6.4/html-single/Administration_and_Configuration_Guide/index.html#chap-Web_Subsystem"/>
                         <link href="http://docs.jboss.org/jbossweb/7.0.x/" title="JBoss Web configuration"/>
                     </hint>
                 </iteration>
+                <iteration over="1">
+                    <when>
+                        <not>
+                            <has-hint />
```
jsight: I don't understand what this is supposed to do?

OndraZizka: It's supposed to prevent adding the classification to a file where the <iteration over="2"> already put a hint.

rules-reviewed/eap6/jboss-eap5/jboss-eap5-xml.windup.xml
OndraZizka: That's right, for these two rules it will prevent duplication. But not for the others, if there are any.
I mean - yes, let's do it that way here, but for general duplication prevention, skipping things touched by other rules by adding <not> could lead to quite complicated conditions. (Therefore my attempt in WINDUP-1050)

OndraZizka: I've tried few combinations, last one this:

```xml
            <when>
                <and>
                    <xmlfile matches="//*[local-name()='jboss-app']" in="jboss-app.xml" as="1"/>
                </and>
                <not>
                    <xmlfile from="1" public-id=".*JBoss.+DTD Java EE.+5.*"/>
                </not>
            </when>
```
But that doesn't work. Any better idea?

OndraZizka: BTW, since we are pretending that we apply the filter to each file, one by one, this *should* work too:

```xml
            <when>
                <not>
                    <xmlfile public-id=".*JBoss.+DTD Java EE.+5.*"/>
                </not>
                 <xmlfile matches="//*[local-name()='jboss-app']" in="jboss-app.xml"/>
            </when>
```
I'm not saying it should work with our current implementation; I'm saying that as a user, I'd intuitively expect this to work.

jsight: We aren't pretending that we apply the filter to each file. We explicitly do not do that at the moment actually.

OndraZizka:
I've removed rule 19000:
Moved XSLT to rule 08000, and removed the "Missing DTD" classification which I don't think has any real value.

rules-reviewed/eap6/websphere/websphere-xml.windup.xml
jsight: Isn't this effectively just the equivalent of removing the classification completely? Since the previous iteration is adding a hint?

OndraZizka: The iteration which adds hint is root element based,
the iteration adding classification is based on file name.
I am not 100 % sure how it really works but I assume these are 2 independent queries, and therefore the 2nd iteration, over "1", can still add classifications to files that are named like that, but do not have the matching root element?

jsight: I think that you need to test this. The FileLocationModel that comes out of the File condition is going to be different from what comes out of the XmlFile condition, so I think that the "not" filter here is going to effectively be a noop.

OndraZizka: Ok I'll do a test.

OndraZizka: You were right :)  See WINDUP-1059 and https://github.com/windup/windup/pull/947 .

jsight:
Merged: https://github.com/windup/windup-rulesets/commit/4aff1198a36023d7aae65f68acb4a22b1e3d49fa

PR #151, opened 2016-05-30
WINDUPRULE-151

mareknovotny:
this depends on https://github.com/windup/windup/pull/927
https://issues.jboss.org/browse/WINDUPRULE-151

### DIFF
--- a/rules-reviewed/eap7/eap6/hsearch.windup.xml
+++ b/rules-reviewed/eap7/eap6/hsearch.windup.xml
@@ -330,6 +330,8 @@
         <rule id="hsearch-00110">
             <when>
                 <javaclass references="org.hibernate.search.backend.configuration.impl.IndexWriterSetting.{constant}">
+                    <location>ENUM_CONSTANT</location>
+                    <location>IMPORT</location>
                 </javaclass>
             </when>
             <perform>
--- a/rules-reviewed/eap7/eap6/tests/data/data-hsearch/HsearchUtil.java
+++ b/rules-reviewed/eap7/eap6/tests/data/data-hsearch/HsearchUtil.java
@@ -88,7 +88,12 @@ public class HsearchUtil {
 
     public void main(String[] args) {
         
-        FullTextSession fullTextSession = Search.getFullTextSession( s );
+        Session session = sessionFactory.openSession();
+        FullTextSession fullTextSession = Search.getFullTextSession( session );
+        
+        FullTextQuery fullTextQuery = fullTextSession.createFullTextQuery( luceneQuery );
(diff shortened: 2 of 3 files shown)


rules-reviewed/eap7/eap6/hsearch.windup.xml
```diff
@@ -330,6 +330,8 @@
         <rule id="hsearch-00110">
             <when>
                 <javaclass references="org.hibernate.search.backend.configuration.impl.IndexWriterSetting.{constant}">
+                    <location>ENUM_CONSTANT</location>
+                    <location>IMPORT</location>
```
Ladicek: What would happen if these two lines weren't added?

mareknovotny: it just searches all type reference locations. see http://windup.github.io/windup/docs/latest/javadoc/org/jboss/windup/ast/java/data/TypeReferenceLocation.html

Ladicek: I mean, I just wondered why are the lines added. No big deal :-)

rules-reviewed/eap7/eap6/tests/data/data-hsearch/HsearchUtil.java
```diff
@@ -127,7 +131,12 @@ public void main(String[] args) {
         BuildContext context = (BuildContext) searchIntegrator;
         context.getIndexingStrategy();
         
-        System.out.println(IndexWriterSetting.TERM_INDEX_INTERVAL + " test");
```
mareknovotny: btw @Ladicek this doesn't work at all, because of type argument in {{System.out.println}} method is {{java.lang.String}} type and that doesn't trigger the analysis for types in case of Enum or Enum constant  while Windup parses with AST.

rules-reviewed/eap7/eap6/hsearch.windup.xml
mareknovotny: just limiting the search ;)

it should find all enumeration constants in Constructor calls, method invocations if the argument is the Enum type and in Variable/Field declarations.

I certainly can keep the rule without the definition of location, but this is just nice example how to use it and don't affect the result IMHO.

rules-reviewed/eap7/eap6/tests/data/data-hsearch/HsearchUtil.java
Ladicek: OK, that's somewhat understandable, because for this to work, Windup would have to have a fully annotated AST, for which it would have to have a complete compilation classpath. I'm not sure how much name and type resolution Windup currently does.

Anyway, I think this doesn't affect us much.

PR #154, opened 2016-06-03
Formatted few rules (wrong indentation).

OndraZizka:
Changed few messages.


jsight:
The tests do not pass and in some of the cases, markdown formatting was changed in ways that make the resulting output less clear to look at (for example, the PostgreSQL JDBC URL rule removed some useful indentation).

OndraZizka:
Does Markdown work if there are spaces before the whole ``` code block?

OndraZizka:
Does Markdown work if there are spaces before the whole ``` code block?

jsight:
Our xml loader deals with that as it is common for the entire text block to be indented in the xml.

d-s:
Let's wait to merge this until after we tag 2.6.0.Final since we don't have very thorough QA right now. (I'm afraid of a big formatting fix like this introducing some issue we don't catch if it's thrown in at the last minute.)

windupgithubbot1:
Build triggered. sha1 is original commit.

mareknovotny:
this needs a rebase if it is still considered for merging.

PR #155, opened 2016-06-03
an experiment with stripped jars produced by class-body-stripper

mareknovotny:


mareknovotny:
@jsight, @OndraZizka are you fine with this reduction or could we get better result ? WDYT?

jsight:
Sorry... I meant to ask about this yesterday. I'm assuming that you tested it and it worked ok? If so, this seems like a good step. +1

mareknovotny:
I didn't do any complete tests on everything, just windup-rulesets tests themselves, some basic apps, like our distribution/samples/ dir and it worked without issue.

jsight:
Ok... merged. If anything comes up we can revert it pretty easily. I suspect that it is ok if the rule tests are passing, though.

PR #158, opened 2016-06-06
Simplified the rule to prevent parameterization from trying to coordi…

jsight:
still testing...


d-s:
lgtm assuming your tests behave as expected

src/test/java/org/jboss/windup/rules/tests/WindupRulesTest.java
```diff
@@ -319,6 +319,7 @@ private void runWindup(GraphContext context, File baseRuleDirectory, final List<
         windupConfiguration.setOutputDirectory(output.toPath());
         windupConfiguration.addDefaultUserRulesDirectory(baseRuleDirectory.toPath());
         windupConfiguration.setOptionValue(SourceModeOption.NAME, sourceMode);
+        windupConfiguration.setOffline(true);
```
jsight: Unrelated change, but this keeps windup's xml validation from triggering on our test data. That is really unnecessary and sometimes slow.

Obviously, the build-time validation against the rules schema still happens, and that is the important part.

PR #160, opened 2016-06-08
Remove unused rules directory

jsight:


mareknovotny:
Are we not going to move from there some rules which were not reviewed and moved to rules-reviewed? I am afraid that some are still outstanding.

jsight:
If we need to do that later we can pull this from the history. Having said that, I think that the remaining rules are completely outdated at this point and it seems unlikely that we would still want them.

PR #161, opened 2016-06-08
Remove effort from catchall rules.

d-s:
When a catchall rule triggers in an application, it tends to trigger many times. This makes catchall rules with effort tend to dominate story point counts in a way that makes them less useful. It also causes performance problems on large applications by attaching effort to many thousands of issues.

Command used to check for instances of catchalls with effort:
d^@^:windup-rulesets$ find rules-reviewed -name "*.windup.xml" | grep catchall | xargs grep effort | grep -v "effort=\"0\""
rules-reviewed/eap6/websphere/websphere-catchall.windup.xml:                        <hint title="IBM proprietary type reference" effort="1" severity="potential">
rules-reviewed/eap6/websphere/websphere-catchall.windup.xml:                        <hint title="IBM ILog proprietary type reference" effort="1" severity="potential">
rules-reviewed/eap6/weblogic/weblogic-catchall.windup.xml:                        <hint title="Oracle proprietary JDBC type reference" effort="1" severity="optional">
rules-reviewed/eap6/weblogic/weblogic-catchall.windup.xml:                        <hint title="WebLogic proprietary type reference" effort="1" severity="potential">

### DIFF
--- a/rules-reviewed/eap6/weblogic/weblogic-catchall.windup.xml
+++ b/rules-reviewed/eap6/weblogic/weblogic-catchall.windup.xml
@@ -201,7 +201,7 @@
                         </not>
                     </when>
                     <perform>
-                        <hint title="Oracle proprietary JDBC type reference" effort="1" severity="optional">
+                        <hint title="Oracle proprietary JDBC type reference" effort="0" severity="optional">
                             <message>
                               <![CDATA[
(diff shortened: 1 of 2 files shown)


jsight:
To put this in perspective, it is rare to see a scrum team do more than ~500 points in a two week sprint. When we add 298k points of effort to a project (as one of these rules did) we seem to be indicating that the migration would take 40 years. This does not seem like a useful way to assess effort.

PR #165, opened 2016-06-27
Updated rules to match the new task 177 and 178 approach to versions

jsight:
Related PR: https://github.com/windup/windup/pull/993

### DIFF
--- a/rules-reviewed/eap6/commonj/commonj.windup.xml
+++ b/rules-reviewed/eap6/commonj/commonj.windup.xml
@@ -12,7 +12,7 @@
       </dependencies>
       <sourceTechnology id="weblogic" />
       <sourceTechnology id="websphere" />
-      <targetTechnology id="eap" versionRange="[6,)" />
+      <targetTechnology id="eap" versionRange="[6,7)" />
       <targetTechnology id="java-ee" versionRange="[6,)" />
       <tag>reviewed-2015-09-25</tag>
    </metadata>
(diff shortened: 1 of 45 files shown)


rules-reviewed/eap6/commonj/commonj.windup.xml
```diff
@@ -12,7 +12,7 @@
       </dependencies>
       <sourceTechnology id="weblogic" />
       <sourceTechnology id="websphere" />
-      <targetTechnology id="eap" versionRange="[6,)" />
+      <targetTechnology id="eap" versionRange="[6,7)" />
```
mareknovotny: so open range won't be supported?

mareknovotny:
I think we need to update also maven-jar-plugin configuration to include '*.windup.technologytransformer.xml' files into the windup-rulesets jar artifact. Somewhere here https://github.com/windup/windup-rulesets/blob/master/pom.xml#L253-L257

jsight:
Good point. I will take a look when I get back.

rules-reviewed/eap6/commonj/commonj.windup.xml
mareknovotny: indeed, I just commented the first change as I noticed you changed all similar rules version ranges ;)

jsight:
@mareknovotny Is this ok to merge as well?

mareknovotny:
yes, I was only disrupted by the mtg we had yesterday ;)

jsight:
Oh, I can understand that.

PR #171, opened 2016-09-02
first draft of quickfixes in some rules triggered in our jee-example-app

mareknovotny:
https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/249

### DIFF
--- a/rules-reviewed/eap6/weblogic/weblogic.windup.xml
+++ b/rules-reviewed/eap6/weblogic/weblogic.windup.xml
@@ -233,6 +233,10 @@
                     <link href="http://www.jboss.org/quickstarts/eap/logging/" title="JBoss Logging Quickstart" />
                     <tag>logging</tag>
                     <tag>weblogic</tag>
+                    <quickfix name="Use a standard logger" type="REPLACE">
+                        <replacement>java.util.logging.Logger</replacement>
+                        <search>weblogic.i18n.logging.NonCatalogLogger</search>
+                    </quickfix>
                 </hint>
             </perform>
         </rule>
--- a/rules-reviewed/eap6/weblogic/weblogic.windup.xml
+++ b/rules-reviewed/eap6/weblogic/weblogic.windup.xml
@@ -233,10 +233,6 @@
                     <link href="http://www.jboss.org/quickstarts/eap/logging/" title="JBoss Logging Quickstart" />
                     <tag>logging</tag>
                     <tag>weblogic</tag>
-                    <quickfix name="Use a standard logger" type="REPLACE">
-                        <replacement>java.util.logging.Logger</replacement>
-                        <search>weblogic.i18n.logging.NonCatalogLogger</search>
-                    </quickfix>
                 </hint>
             </perform>
         </rule>
--- a/rules-reviewed/droolsjbpm/kie-api.windup.xml
+++ b/rules-reviewed/droolsjbpm/kie-api.windup.xml
@@ -41,6 +41,10 @@
          <perform>
             <hint title="org.drools.event.AgendaEventListener is deprecated" effort="0">
                <message>Replace with : org.kie.api.event.rule.AgendaEventListener</message>
+               <quickfix name="AgendaEventListener-qf1" type="REPLACE">
+                    <replacement>org.kie.api.event.rule.AgendaEventListener</replacement>
+                    <search>org.drools.event.AgendaEventListener</search>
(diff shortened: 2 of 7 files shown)


mareknovotny:
it seems artifactory is again on strike to serve an artifact
retest, please!

mareknovotny:
retest

rules-reviewed/eap6/weblogic/weblogic.windup.xml
```diff
@@ -106,6 +106,22 @@
                     <message>Replace with the Java EE standard method `javax.transaction.TransactionManager.resume(Transaction tx)`.</message>
                     <tag>transactions</tag>
                     <tag>weblogic</tag>
+                    <quickfix name="Use java ee standard method Transaction Manager for resuming 1" type="REPLACE">
+                        <replacement>javax.transaction.TransactionManager.resume</replacement>
+                        <search>weblogic.transaction.TransactionManager.resume</search>
```
d-s: I thought <search>/<replacement> would specify a string replacement? So I don't see how these would work -- the line the rule triggers on will look something like:

ctm.forceResume(saveTx);

The way these are written implies that the code processing the quickfix would be recognizing a reference to a method call from a particular class, and then replacing it with a call to a method from another class.

The first part (recognizing the reference) we can do, but we would need to have a way to identify whether this quickfix should refer to a string replacement or a reference lookup, right? And the replacement is really not simple at all, because you can't change the type of object the method is being invoked on without finding where that object was created and creating it differently...

I'm guessing for this rule any quickfix would be difficult. Probably we would handle it by offering a quickfix for another rule like the one that triggers on the line where the WebLogic TransactionManager is constructed.

mareknovotny: right, I was switched to something else in my head and I forget on that it is pure string not AST ref.

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
retest

jsight:
It looks like the testing failures are because of a JBoss Nexus failure. Basically the load balancer is bouncing between multiple servers and at least one is out of sync. So Artifactory gets intermittent 404 errors from it. I can reproduce it here with wget as well.

There is an engineering ticket for it, so hopefully it will get sorted out soon.

mareknovotny:
ok, thanks @jsight for info

mareknovotny:
retest

mareknovotny:
for some reason it downloads not the latest windup-config-xml snapshot and due this it fails on xml validation.

jsight:
That's probably JBoss Nexus serving stale metadata or something.

mareknovotny:
I checked the metadata from jboss.org and it seems correct, the latest snapshot is 3.0.0-20160902.105500-32 where artifactory serves thisI will give a time to

jsight:
retest

jsight:
Hmmm, seems they fixed the caching issue with Nexus but we have a stale version:
http://wonka.mw.lab.eng.bos.redhat.com/artifactory/jboss/org/jboss/windup/config/windup-config-xml/3.0.0-SNAPSHOT/maven-metadata.xml

Will check...

jsight:
retest

d-s:
@mareknovotny @jsight We should perhaps move ahead with getting these PRs reviewed/submitted regardless of the PR builder status, since stuff on the Eclipse side will start to stack up behind them.

jsight:
@d-s - The build passes, I thought that the main question was whether or not the rules were sufficient?

It seems like the first one would be very unlikely to match in most cases.

d-s:
@jsight Yeah, the first one has the same problem the other rule originally in the PR had.
@mareknovotny Can you see about finding another quick rule or two we can fix with a pure string replace?

mareknovotny:
@jsight @d-s we need to move on as it seems, this is blocking @johnsteele in his work. Could you review the change?

jsight:
I will go ahead and merge it so that we can see how this works in Eclipse. I am still a little concerned that this is going to produce quite a few quick fixes that really won't do what the user expects.

PR #172, opened 2016-09-07
Various improvements, fixes and extensions

Maarc:
This PR bundles a couple of changes I did on my local rule set.

It comprises changes like:
- title normalization (for example to have all hibernate related issues starting with "Hibernate")
- formatting
- level of effort updated (0->1 and 1 -> 3)
- severity update (mandatory -> optional)
- new rules

PR #173, opened 2016-09-12
Maarc changes polished

mareknovotny:
This PR is enhanced PR from https://github.com/windup/windup-rulesets/pull/172

Not wanted changes are missing and there are accompanying tests changes too.

### DIFF
--- a/rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
+++ b/rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
@@ -23,7 +23,7 @@
                 </javaclass>
             </when>
             <perform>
-                 <hint title="Dynamic class instantiation" effort="1" severity="optional">
+                 <hint title="Dynamic class instantiation" effort="0" severity="optional">
                      <message>
                        <![CDATA[
                        The class is dynamically loaded within application. During the migration, multiple classes that are provided on classpath by a different server may not be present anymore.
--- a/rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
+++ b/rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
@@ -23,7 +23,7 @@
                 </javaclass>
             </when>
             <perform>
-                 <hint title="Dynamic class instantiation" effort="0" severity="optional">
+                 <hint title="Dynamic class instantiation" effort="1" severity="optional">
                      <message>
                        <![CDATA[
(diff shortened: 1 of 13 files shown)


rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
```diff
@@ -23,7 +23,7 @@
                 </javaclass>
             </when>
             <perform>
-                 <hint title="Dynamic class instantiation" effort="0" severity="optional">
+                 <hint title="Dynamic class instantiation" effort="1" severity="optional">
```
d-s: We decided in the past to keep this at 0 -- if effort is attached this rule often completely takes over the story point estimates for the entire app.

rules-reviewed/eap6/websphere/websphere-catchall.windup.xml
```diff
@@ -17,6 +17,31 @@
...
+            </when>
+            <perform>
+                <iteration>
+                    <when>
+                        <not>
+                            <has-hint />
+                        </not>
+                    </when>
+                    <perform>
+                        <hint title="IBM DB2 driver type reference" effort="1" severity="potential">
```
d-s: no effort on catchalls

mareknovotny: While I certainly agree with this, this particular case makes sense to have effort as that means you need to configure JNDI/Datasource.

rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
mareknovotny: I know, I feel it is sometimes wrong ;)

rules-reviewed/eap6/websphere/websphere-catchall.windup.xml
jsight: True, but you won't likely need to configure it 100 times. This hint, on the other hand, could fire many times in a single application.

d-s: Yeah, my main concern here is making sure that we don't attach effort to rules that will often trigger thousands of times, because it causes performance issues and useless story points numbers.

I'm fine with checking it in with effort first, and then looking at the perfcheck and test apps a few days later to see what the impact is. (I think we need to do this for any large rules PR.)

I do think if a rule needs effort it should be in a different ruleset from the catchalls, but I don't feel strongly about that part.

mareknovotny:
I took in mind all comments and lower the effort to 0 for both cases.

PR #178, opened 2016-11-14
WebSphere rules for EAP7

mareknovotny:
taiga task #350 and #351


rules-reviewed/eap6/websphere/websphere-catchall.windup.xml
```diff
@@ -8,11 +8,11 @@
...
         </description>
         <dependencies>
-            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.3.0.Final" />
-            <addon id="org.jboss.windup.rules,windup-rules-java,2.3.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.7.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.7.0.Final" />
         </dependencies>
         <sourceTechnology id="websphere" />
-        <targetTechnology id="eap" versionRange="[6,7)" />
+        <targetTechnology id="eap" versionRange="[6,7]" />
```
jsight: Do you think that the links in this could be updated to point to an EAP 7 equivalent doc?

rules-reviewed/eap6/websphere/websphere-xml.windup.xml
```diff
@@ -6,11 +6,11 @@
             This ruleset provides analysis of WebSphere proprietary xml files and constructs that may require individual attention when migrating to JBoss EAP 6+.
         </description>
         <dependencies>
-            <addon id="org.jboss.windup.rules,windup-rules-xml,2.3.0.Final"/>
+            <addon id="org.jboss.windup.rules,windup-rules-xml,2.7.0.Final"/>
         </dependencies>
         <sourceTechnology id="websphere"/>
-        <targetTechnology id="eap" versionRange="[6,7)"/>
-        <tag>reviewed-2015-06-17</tag>
+        <targetTechnology id="eap" versionRange="[6,7]"/>
```
jsight: Could you clone this into a new ruleset instead of reusing the existing one? I am concerned that the rules are a little unclear for the user if they have both EAP6 and EAP7 links in them.

rules-reviewed/eap6/websphere/websphere-catchall.windup.xml
jsight: @d-s also pointed out that this should be [6,8) in order to cover minor versions (for example, 7.1). I think this will need to be changed as well.

mareknovotny: @jsight Links are going to be updated I create jira issues for this and asked  @sgilda to do that and she did the updates AFAIK for some and others will be filed or updated if needed.

I wasn't sure with the updating to 6 - 8 version range, but I think you both are right, so I will update them to that version range.

rules-reviewed/eap6/websphere/websphere-xml.windup.xml
mareknovotny: I would keep this rather in one ruleset file as the suggestion for both are the same and I included specific links to 6 or 7 for reference as user can use what they needed for his concrete migration. This would reduce the maintenance later I think.

jsight:
@d-s @mareknovotny - I think that the changes themselves are fine, although I am still a little worried about mixing the 7 rules with the existing 6 rules. Perhaps we can have a discussion about it in the standup tomorrow?

mareknovotny:
ok, I am ok for discussing that issue on standup call today.

mareknovotny:
I did a split of websphere-xml.windup.xml rules as we agreed and I kept catchall/ignored rules for 6/7.

mareknovotny:
retest

jsight:
retest

jsight:
retest

PR #185, opened 2017-01-16
Effort increased for the valve rule

Maarc:

PR #186, opened 2017-01-16
Minor hint, effort and title corrections

Maarc:


rules-reviewed/eap7/eap5/jboss-eap5-7-java.windup.xml
```diff
@@ -208,7 +209,7 @@
             </when>
             <perform>
                 <iteration>
-                    <classification title="JBoss Web Valve" effort="1" category-id="mandatory">
+                    <classification title="JBoss Web Valve" effort="3" category-id="mandatory">
```
d-s: This change belongs with PR #185?

windupgithubbot1:

rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
```diff
@@ -23,7 +23,7 @@
                 </javaclass>
             </when>
             <perform>
-                 <hint title="Dynamic class instantiation" effort="0" category-id="optional">
+                 <hint title="Dynamic class instantiation" effort="1" category-id="optional">
```
d-s: We've seen this change proposed several times, and can't merge it as things stand -- on large apps it can trigger thousands of times, and make up >90% of the story point estimate. We can reopen the topic of how to handle issues like this (like perhaps by having options for effort scaling other than [constant * #incidents]) but until then we can't attach effort to this hint.

Maarc: This is an "optional" change. The real estimate for a migration is only focussing on the "mandatory" category. "Optional" efforts are relevant for a modernization/optimization/cleanup.

An intense usage of reflection and dynamic class instantiations "smells". Cleaning it up properly definitely will take a while. Here we have an indication that something could be done here.

d-s: These are reasonable points, but:

-- We report a top-line story points number (in large font on App List and App Details), and this count includes effort for all issues, not just mandatory ones. Effort counts from different categories of issues are also aggregated in other places, like the per-jar sections on App Details. Maybe the reports should not do this, and should highlight effort counts that only take into account for mandatory issues? But for now that's how things are counted.

-- Story point estimates dominated by many appearances of one issue type are confusing, as they imply that >90% of the migration effort comes from that one type of issue, and drown out the other types of issues present in the app.

We really should make some changes to the way effort is counted and reported, and this should probably be a focus in the next release cycle after the initial release of windup-web. Until then, though, we really can't merge this change. It will change story point estimates to something that is very different from previous versions of Windup and confusing to most users.

My suggestion for rules like this, where you've proposed a change but we couldn't merge it, is to use the rules override feature that was introduced in Windup 2.5. This will let you maintain a personal version of the rules that overrides some of our decisions not to incorporate some changes into the core rulesets and keeps your changes intact as you pull in updates to the basic rules.

mareknovotny: Well, it seems to like it is more than Optional as Potential issue category-id from this and previous discussions. Sometimes there it is a problem, it could have higher effort than 1 anyway.
Btw, everybody can take an existent rule and overwrite (copy it into your own location and keep the id) its properties (effort, titles, etc). We have this feature from 2.6.0.Final - 
look at https://access.redhat.com/documentation/en/red-hat-jboss-migration-toolkit/2.7/paged/windup-rules-development-guide/chapter-5-overriding-rules

mareknovotny:
This looks good to me except the raising of effort in environment-dependent-calls-01000 rule. What we do with that?

mareknovotny:
retest

rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
Maarc: @d-s  I agree that it has an impact on aggregated efforts. 

Thanks for the hint on the override feature. I will then override the classloading rule from my side and we should discuss the effort aggregation later.

Maarc:
Is there something preventing to merge this rule with the master branch?

mareknovotny:
@Maarc yes, first raising effort in **environment-dependent-calls-01000** rule -> https://github.com/windup/windup-rulesets/pull/186#issuecomment-273108352 and second I would asked for a rebase of your branch on this PR

rules-reviewed/eap6/uncategorized/environment-dependent.windup.xml
OndraZizka: I am not sure what's the @Maarc 's current wish. Should it be 1 or 0? Do the PR's here represent the changes you want?

OndraZizka:
Merged with the exception of environment-dependent-calls-01000 which is still 0 due to the reasons discussed.

PR #187, opened 2017-01-16
Adjustments to the WebSphere rules

Maarc:
Removed classification leading to duplicated entries. Corrected efforts for catchalls from 0 to 1.


windupgithubbot1:

mareknovotny:
classification and hint is not a duplication generally, and we did the opposite effort change in autumn https://github.com/windup/windup-rulesets/commit/f33bcd5e809d3099ef2444b31825891e6e798d72

I think this PR needs to be discussed with @d-s, @jsight first if we would like to do a reverse change

rules-reviewed/eap6/websphere/websphere-catchall.windup.xml
```diff
@@ -83,7 +83,7 @@
                         </not>
                     </when>
                     <perform>
-                        <hint title="IBM ILog proprietary type reference" effort="0" category-id="potential">
+                        <hint title="IBM ILog proprietary type reference" effort="1" category-id="potential">
```
Maarc: Let's leave "0" for the moment as discussed in our conf-call several weeks ago.

In the future, it needs to be clear that all issues flagged as "potential". As such they potentially encompass the highest efforts and risks.

Maarc:
The hint and classification here are producing duplicates in the generated reports. The classification should be removed.

mareknovotny:
@Maarc please change the effort back as we discussed above and I will merge it.

rules-reviewed/eap6/websphere/websphere-catchall.windup.xml
OndraZizka: @Maarc  What do you mean by "In the future, it needs to be clear that all issues flagged as "potential". As such they potentially encompass the highest efforts and risks." ?

OndraZizka: IMO, I think the potential's should be left to the users to customize. Windup is a rule based toolkit exactly for purpose - so that people can customize what it does. Someone will want 0, someone 32.

OndraZizka: I am leaving 0 at all. Setting them to 1 leaded to an explosion of the story points. Maybe we should generally shift the whole scale so that 1000x 1sp issue which you can handle with a regexp replace wouldn't beat an issue with 64sp which is for several weeks of coding.

OndraZizka:
Leaving the IBM DB2 driver type reference at effort 1. The others would lead to SP go explode

PR #200, opened 2017-05-04
T840 Remove duplicated hint

OndraZizka:


mareknovotny:
We talked about it on mtg, that we need to change migration path to have source target(s) for common migrations from Anything to EAP 7/6. So closing this PR

jsight:
Well, I don't think that they will have sources and targets for those. I think these rules just need to be refactored so that the completely identical ones are in a common location.

PR #209, opened 2017-06-01
Deduplicate rules for EAP5 and 6 to 7 based upon feedback

jsight:


mareknovotny:
just note for myself - this moves some rules from EAP5-7 rules into EAPX-7 rules to not trigger them both when sourceTechnology is not spefified in Analysis (execution) configuration.

PR #214, opened 2017-06-06
Deduplicate websphere-to-eap6or7

OndraZizka:


rules-reviewed/README.txt
```diff
@@ -0,0 +1,14 @@
+This directory contains default set of Windup rules distributed with RHAMT.
```
jsight: Good idea... although maybe once this current round of fixes is done, we can fix the name. rules-reviewed shouldn't stay here forever, IMO.

jsight:
-1 on deduplication in this way. This duplicates a lot of links and isn't really necessary.

I think the duplication in the slide should only focus on the duplication between classification and hint.

rules-reviewed/README.txt
OndraZizka: haha

OndraZizka:
Are you sure? I can barely read the screenshot but I see one entry for EAP 7 and one for EAP 6.

jsight:
@OndraZizka - I think we need a separate PR to insure that the correct target is run (eap6 or 7 vs just "eap").

PR #220, opened 2017-06-12
WINDUPRULE-227

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-227


jsight:
Looks good and it seems like this caught all of them.

PR #221, opened 2017-06-13
WR-219 Remove the info classification

OndraZizka:


rules-reviewed/eap7/weblogic/weblogic-xml-descriptors.windup.xml
```diff
@@ -88,7 +88,9 @@
                 </or>
             </when>
             <perform>
+                <!--
                 <classification title="WebLogic EJB XML" category-id="mandatory" effort="0" />
+                -->
```
mareknovotny: @OndraZizka  but you are effectively adding that classification or I don't get the comment and PR title ;)

PR #250, opened 2017-07-17
Back-port of changes from EAP 7 to 7. Various fixes and corrections.

Maarc:
Here are some fixes for a couple of EAP 6 rules that have not been updated.

### DIFF
--- a/rules-reviewed/eap6/jboss-eap5/jboss-eap5-xml.windup.xml
+++ b/rules-reviewed/eap6/jboss-eap5/jboss-eap5-xml.windup.xml
@@ -81,7 +81,7 @@
                 </xmlfile>
             </when>
             <perform>
-                <classification issue-display-mode="all" title="JBoss Seam Components" effort="5">
+                <classification issue-display-mode="all" title="JBoss Seam Components (components.xml)" effort="5">
                     <description>
                         Allows to configure a Seam component using XML: to isolate deployment-specific information from
                         the Java code, to enable the creation of re-usable frameworks, to configure Seam's built-in functionality, etc.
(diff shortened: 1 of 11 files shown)


mareknovotny:
ok, thanks @Maarc , but I can see there are also EAP7 rules changed so it is mixed. We can't merge it as it is.

jsight:
@mareknovotny - I do not think that we can expect Marc to do things in a clean way. IMO, we should review this and make the test changes to get it merged.

I am concerned about the links to old knowledgebase articles, though. Is there any way to confirm that these will work with EAP 7? If they do, we should put in the edits to get the EAP 7 added to the list of applicable targets for the KB articles. @bergerhoffer, can you help with that part?

mareknovotny:
@jsight yep, I am fine to pull some of changes from this PR commit.

The re-adding of KB links for EAP6 to EAP7 though is a little controversial. First even the content is correct and valid for EAP7 not mentioning EAP7 there could readers confused. Second I am not really sure as I read comments before I created that https://issues.jboss.org/browse/WINDUP-1242 an all others jiras related to not updated EAP7 links.

jsight:
@mareknovotny, @Maarc - IMO, the knowledgebase content all has to be tested. We shouldn't link to it if it doesn't actually work and we should know whether it works.

This puts them into two categories:

 - The ones that don't work (we shouldn't link to these)
 - The ones that do work -- it shouldn't take more than a day or so to get EAP 7 added to the knowledgebase article itself, if we've tested them and they do work, eh?

I don't think that we should link to them if we don't have a solid expectation for them to actually work.

Maarc:
@mareknovotny @jsight 

I have seen linked EAP 6 KB articles in EAP 7 rules. In this PR I also updated some article titles to have the same than on other existing rules.

As long as there is no article for EAP 7, in my opinion, it is better to display the link to the existing EAP 6 articles and mark it clearly as unverified for EAP 7. While trying to solve such issues, it is better to provide all available information, even if it is unverified. I am expecting 90% of the EAP 6 KB articles to work with EAP 7.

In the cases where we are sure that the EAP 6 article will / cannot work at all for EAP 7, it is a good idea to remove the link.

mareknovotny:
I kept the changed link titles, but not the problematic links. see PR #251 what is there.

PR #255, opened 2017-07-26
WINDUPRULE-266: new Seam API rules

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-266

### DIFF
--- a/rules-reviewed/eap6/java-ee/seam/seam-java.windup.xml
+++ b/rules-reviewed/eap6/java-ee/seam/seam-java.windup.xml
@@ -24,13 +24,17 @@
             <perform>
                 <classification title="Seam API" effort="5" category-id="mandatory">
                     <description>
-                        Seam 2.2 and earlier is not supported on JBoss EAP 6 and above. Consider migration to Context Dependency Injection (CDI) standard which covers most of Seam 2 core functionalities in
+                        Seam 2.2 and earlier is not supported on JBoss EAP 6 and above.
+                        Consider migration to Context Dependency Injection (CDI) standard which covers most of Seam 2 core functionalities in
                         a standardized, type safe and extensible way.
                         
+                        Seam 2.3 still could work on EAP 6.x, but is not maintained and officially supported on new EAP 6.x patched releases or EAP 7.x.
+
                         There are two options available:
                         
-                        1. Continue using Seam 2.x. This approach is sometimes lower effort but the application will not use a tested and supported library and it may not work on EAP 7 at all.
-                        2. Switch to standard CDI beans and migrate to JSF 2.2. This will require significant migration effort.
+                        1. Continue using Seam 2.x on EAP 6 but expect there are bugs or glitches and you need to fix it yourself.
+                        This approach is sometimes lower effort but the application will not use a tested and supported library.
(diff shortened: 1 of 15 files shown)


jsight:
I am a little bit concerned that the guidance for migration options is not really sufficient. If I understand it correctly, there is really a matrix of scenarios:

<table>
<tr>
<td>Seam Version</td>
<td>EAP 6</td>
<td>EAP 7</td>
<tr>
  <td>Seam 2.2</td>
  <td>Compatible but unsupported - requires switching to the JSF 1.x slot</td>
  <td>Not compatible. May be possible to get it working, but would require extensive changes that are not documented.</td>
</tr>
<tr>
  <td>Seam 2.3</td>
  <td>Compatible but unsupported</td>
  <td>Compatible but unsupported</td>
</tr>
</table>

I think that we should give the user all of these details in the rules that provide the user with a summary of their options. Please update the text if the guidance that I wrote above is incorrect, though.

mareknovotny:
The table matrix was not completely correct, see this: 
<table>
<tr>
<td>Seam Version</td>
<td>EAP 6</td>
<td>EAP 7</td>
<tr>
  <td>Seam 2.2</td>
  <td>Compatible but unsupported - requires switching to the JSF 1.x and Hibernate 3 slots</td>
  <td>Not compatible. May be possible to get it working, but would require extensive changes that are not documented.</td>
</tr>
<tr>
  <td>Seam 2.3</td>
  <td>Compatible but unsupported and not tested on current EAP 6.x</td>
  <td>Not compatible</td>
</tr>
</table>

I thought that I give such guidance in [Seam API classification](https://github.com/mareknovotny/windup-rulesets/blob/WINDUPRULE-266/rules-reviewed/eap6/java-ee/seam/seam-java.windup.xml#L25-L51) , but if you think it is ambiguous or some unclarity, I can add the table matrix too. WDYT?

jsight:
@mareknovotny - The problem with the description in the rule now is that it implies that they can continue to use Seam 2.x with EAP 7. At the time, I thought that might be possible and that is what Marc was demanding. I have since tried some apps and found that it was not as simple as just keeping it, and that seems to be what your updated table says as well.

I agree with your updates, with one caveat. It seems to me that Seam 2.3 did work on EAP 7 in the one (small) test with a quickstart that I did there. I'm sure there are some broken features, though, and obviously we have no documentation to say what works and what doesn't.

jsight:
@mareknovotny - The same could be said for trying to run JSF 1.2 in it generally, but we tell them that it might work. ¯\_(ツ)_/¯ 

Basically, it isn't recommended and if it breaks, they get to keep the pieces, but it would be consistent with the other "advice" that we have been required to make (regarding hibernate and JSF).

mareknovotny:
@jsight  so what else in comparison to the latest commit we need to change/add here, I am a little unsure?

jsight:
@mareknovotny - My only suggestion was to clarify the situation regarding Seam 2.3 on EAP 7, as it can be made to work under some circumstances. Basically this could be pretty similar to the guidance for Seam 2.3 on EAP 6, in that it can be made to work, but isn't recommended.

I can show you the app that I ported over to EAP 7 if you'd like? It was surprisingly easy, but of course it wasn't an extensive test (only a few pageflows and some basic templates).

I could not get JSF 1.2 to work at all and thus could not get anything from Seam 2.2 to work in EAP 7.

PR #258, opened 2017-08-23
WINDUPRULE-278: Fixing cloud readiness rules

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-278

I haven't included the following 2 changes for javaclass references:

- https://github.com/Maarc/windup-rulesets/commit/34e22f864ec30eeed35c67606da77f17005cd9a9#diff-68179b4994e8c1c844d59163c749a893L193
- https://github.com/Maarc/windup-rulesets/commit/34e22f864ec30eeed35c67606da77f17005cd9a9#diff-096466b775ea2df36893bdce01c54584L23

Does he have some reasoning for them?

### DIFF
--- a/rules-reviewed/openshift/local-storage.windup.xml
+++ b/rules-reviewed/openshift/local-storage.windup.xml
@@ -30,11 +30,11 @@
                     </javaclass>
                     <javaclass references="java.nio.file.Paths.get({*})">
                         <location>METHOD_CALL</location>
-                    </javaclass>
+                    </javaclass> 
                 </or>
             </when>
             <perform>
-                <hint title="Local file system access" effort="1" category-id="cloud-mandatory">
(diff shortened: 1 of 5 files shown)


jsight:
@mareknovotny - He included those because those rules are really very verbose. I think that we should include them, although I'm not sure that they fully solve the problem.

mareknovotny:
well, it depends on the usage, as `java.nio.file.{*}` could be more verbose as it includes more classes than what we have listed in the condition.
Ok, I will add them, even I think we could use something different from IMPORT location.

jsight:
@mareknovotny - This is good to merge once those other commits are added.

I think we can ask Marco to look at these when he gets back. There really is no great balance to be had here, IMO, as the usage of these APIs really is not wrong.

PR #266, opened 2017-09-13
Winduprule 284: Web technologies (JSF, JSP, etc)

jsight:
Just to demonstrate a few different approaches, as these are a little different. Tests are not yet complete.


mareknovotny:
retest

jsight:
@mareknovotny I don't believe that the tests will pass until the related PR is there.

mareknovotny:
That is fine, I think I would like to just verify that the tests are not in errors, I noticed some weird errors in  the CI PR buider, but I tried more PRs even this was expected to fail related test.

PR #268, opened 2017-09-14
WINDUPRULE-290 - connect technologies identification

klinki:
#close WINDUPRULE-290

### DIFF
--- a/rules-reviewed/technology-usage/connect-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/connect-technology-usage.windup.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0"?>
-<ruleset id="technology-usage-http" xmlns="http://windup.jboss.org/schema/jboss-ruleset" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<ruleset id="technology-usage-connect" xmlns="http://windup.jboss.org/schema/jboss-ruleset" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
     <metadata>
         <description>
-            This ruleset provides statistical summaries of the http items that were found during the analysis.
+            This ruleset provides statistical summaries of the other connect technologies (JCA, JNI, Mail, RMI) items that were found during the analysis.
         </description>
         <dependencies>
             <addon id="org.jboss.windup.rules,windup-rules-javaee,2.7.0.Final" />
--- a/rules-reviewed/technology-usage/connect-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/connect-technology-usage.windup.xml
@@ -1,0 +1,99 @@
+<?xml version="1.0"?>
+<ruleset id="technology-usage-http" xmlns="http://windup.jboss.org/schema/jboss-ruleset" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
+            This ruleset provides statistical summaries of the http items that were found during the analysis.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.7.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.7.0.Final" />
+        </dependencies>
+        <phase>PostMigrationRulesPhase</phase>
+    </metadata>
+    <rules>
+        <rule id="technology-usage-connect-01000">
+            <when>
+                <graph-query discriminator="RMIServiceModel" />
+            </when>
+            <perform>
+                <technology-identified name="RMI">
+                    <tag name="Connect"/>
+                    <tag name="Other"/>
+                    <tag name="Java EE"/>
+                </technology-identified>
+            </perform>
+        </rule>
+
+        <rule id="technology-usage-connect-01100">
+            <when>
+                <or>
+                    <javaclass references="java.lang.System.{systemNativeMethods}({*})">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="java.lang.Runtime.load({*})">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="java.lang.Runtime.loadLibrary({*})">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="com.sun.jna.{*}">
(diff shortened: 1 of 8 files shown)


mareknovotny:
Is this ready for review?

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
```diff
@@ -0,0 +1,99 @@
...
+                    <javaclass references="java.lang.System.{systemNativeMethods}({*})">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="java.lang.Runtime.load({*})">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="java.lang.Runtime.loadLibrary({*})">
+                        <location>METHOD_CALL</location>
+                    </javaclass>
+                    <javaclass references="com.sun.jna.{*}">
```
mrizzi: Good to identify also JNA usage that it's not mentioned in the mock-ups, but i think it should be better to have a rule dedicated to JNA (and so also a `technology-identified` tag for JNA) as JNA has been created to not use JNI ;)
`jni-native-code.windup.xml` rules take them together to address the compatibility of this native code with containers but from the technologies perspective they're two different things.

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
```diff
@@ -0,0 +1,99 @@
...
+                    <tag name="Other"/>
+                    <tag name="Java EE"/>
+                </technology-identified>
+            </perform>
+        </rule>
+
+        <rule id="technology-usage-connect-01100">
+            <when>
+                <or>
+                    <javaclass references="java.lang.System.{systemNativeMethods}({*})">
```
mrizzi: `systemNativeMethods` is not declared so better add:
```xml
            <where param="systemNativeMethods">
                <matches pattern="(load|loadLibrary|mapLibraryName)"/>
            </where>
```

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
```diff
@@ -0,0 +1,99 @@
...
+                </or>
+            </when>
+            <perform>
+                <technology-identified name="JNI">
+                    <tag name="Connect"/>
+                    <tag name="Other"/>
+                    <tag name="Java EE"/>
+                </technology-identified>
+            </perform>
+        </rule>
```
mrizzi: Not a thing that directly apply only to this rule, but i'm asking myself if there would be a better way to handle this kind of rules without duplicating already existing rules.
For example, adding `<technology-tag level="INFORMATIONAL">JNI</technology-tag>` to [jni-native-code.windup.xml](https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/openshift/jni-native-code.windup.xml#L57) would make this rule without all the duplicated conditions in `<when>` section that would become:
```xml
            <when>
                <graph-query discriminator="TechnologyTagModel">
                    <property name="name">JNI</property>
                </graph-query>
            </when>
```

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
```diff
@@ -0,0 +1,99 @@
...
+
+        <rule id="technology-usage-connect-01200">
+            <when>
```
mrizzi: See comment above.
In this case it would be adding `<technology-tag level="INFORMATIONAL">Mail</technology-tag>` in [mail.windup.xml](https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/openshift/mail.windup.xml#L41)

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0"?>
+<ruleset id="technology-usage-http" xmlns="http://windup.jboss.org/schema/jboss-ruleset" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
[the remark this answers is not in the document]
mareknovotny: I would recommend to use id derived from ruleset filename. I noticed it is not always like that. This helps to directly identify the filename.windup.xml

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
```diff
@@ -0,0 +1,99 @@
...
+         xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
+            This ruleset provides statistical summaries of the http items that were found during the analysis.
```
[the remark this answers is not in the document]
mareknovotny: +1

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
klinki: Thanks, I forgot to copy that one

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
klinki: To be honest, in this case I support @OndraZizka suggestion not to write new rules for identifying technologies, but rather update some existing rules and add identification there.

IMHO it would be much easier to maintain. Also it would reduce number of executed rules and possibly improve performance.

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
klinki: Thanks for pointing that out

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
klinki: Btw isn't `jni-native-code.windup.xml` executed only when cloud readiness rules are executed? 
(I don't know if that would be a problem or not.. ?)

mrizzi: For completeness about your comment, you suggest to not write this `technology-usage-connect-01100` rule but rather add the `<technology-identified>` tag to [jni-native-code-00000](https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/openshift/jni-native-code.windup.xml#L39) in order to not duplicate the rule. Right?

klinki: I think that approach might be better. But it was also part of discussion we had with Ondra. (He was suggesting to reuse classification model). Resolution was to stick to current model, so we can ignore that suggestion for now. 

But for future, if we ever get back to it again, I believe it would be better.

klinki:
WRT `<technology-tag level="INFORMATIONAL">` suggestions. I like that idea, but I'm not sure about that, because these rules would be depending on `cloud-readiness` rules, wouldn't they?

(Identification of JNI and Mail and tagging it with `INFORMATIONAL` tag would be done in `cloud-readiness` ruleset... )

rules-reviewed/technology-usage/connect-technology-usage.windup.xml
jsight: It is already possible to reuse the classification information (or technology tags) in the when condition of these rules.

PR #269, opened 2017-09-15
WINDUPRULE-294: Statistics Rules: Security

jsight:


mareknovotny:
retest

rules-reviewed/technology-usage/security-technology-usage.windup.xml
```diff
@@ -0,0 +1,43 @@
...
+                            in="web.xml"/>
+
+                    <xmlfile matches="//*[local-name() = 'login-config']/*[local-name() = 'realm']"
+                            in="{*}ejb-jar.xml"/>
+
+                    <javaclass references="javax.annotation.security.DeclareRoles">
+                        <location>ANNOTATION</location>
+                    </javaclass>
+
+                    <javaclass references="javax.annotation.security.RolesAllowed">
```
mrizzi: I wonder if, from the "Security Realm" perspective, any annotation from `javax.annotation.security` should count as an instance of that technology and not only these two.

mareknovotny: the question is how is the request for tech statistics stated. Is it to count with Realms defined/used or count with usage occurences doesn't matter if 1 or more realms.

jsight: I do not think a count is particularly important. I think it can scan all of the annotations.

jsight: Fixed :)

PR #270, opened 2017-09-18
WINDUPRULE-295: adding JTA technology stats rules

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-295
depends on https://github.com/windup/windup/pull/1179

PR #271, opened 2017-09-20
WINDUPRULE-292 - database

klinki:
Depends on: https://github.com/windup/windup/pull/1201

### DIFF
--- a/rules-reviewed/technology-usage/database-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/database-technology-usage.windup.xml
@@ -24,6 +24,20 @@
                 </technology-identified>
             </perform>
         </rule>
+        <rule id="technology-usage-database-01001">
+            <when>
+                <graph-query discriminator="DataSourceModel">
+                    <property name="isXA">true</property>
+                </graph-query>
+            </when>
+            <perform>
+                <technology-identified name="JDBC XA datasources">
+                    <tag name="Store"/>
+                    <tag name="Database"/>
+                    <tag name="Java EE"/>
+                </technology-identified>
+            </perform>
+        </rule>
         <rule id="technology-usage-database-01100">
             <when>
                 <graph-query discriminator="JPAEntityModel" />
--- a/rules-reviewed/technology-usage/database-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/database-technology-usage.windup.xml
@@ -14,7 +14,7 @@
     <rules>
         <rule id="technology-usage-database-01000">
             <when>
-                <graph-query discriminator="JdbcDatasourceModel"></graph-query>
+                <graph-query discriminator="JdbcDatasourceModel"  />
             </when>
             <perform>
                 <technology-identified name="JDBC datasources">
@@ -26,7 +26,7 @@
         </rule>
         <rule id="technology-usage-database-01100">
             <when>
-                <graph-query discriminator="JPAEntityModel"></graph-query>
+                <graph-query discriminator="JPAEntityModel" />
             </when>
             <perform>
                 <technology-identified name="JPA entities">
@@ -38,7 +38,7 @@
         </rule>
         <rule id="technology-usage-database-01200">
             <when>
-                <graph-query discriminator="JPANamedQueryModel"></graph-query>
+                <graph-query discriminator="JPANamedQueryModel"  />
             </when>
             <perform>
                 <technology-identified name="JPA named queries">
@@ -50,7 +50,7 @@
         </rule>
         <rule id="technology-usage-database-01300">
             <when>
-                <graph-query discriminator="JPAPersistenceUnitModel"></graph-query>
+                <graph-query discriminator="JPAPersistenceUnitModel"  />
             </when>
             <perform>
                 <technology-identified name="Persistence units">
--- a/rules-reviewed/technology-usage/database-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/database-technology-usage.windup.xml
@@ -14,7 +14,9 @@
     <rules>
         <rule id="technology-usage-database-01000">
             <when>
-                <graph-query discriminator="JdbcDatasourceModel"  />
+                <graph-query discriminator="DataSourceModel">
+                    <property name="isXA" type="BOOLEAN">false</property>
+                </graph-query>
             </when>
             <perform>
                 <technology-identified name="JDBC datasources">
@@ -27,7 +29,7 @@
         <rule id="technology-usage-database-01001">
             <when>
                 <graph-query discriminator="DataSourceModel">
-                    <property name="isXA">true</property>
+                    <property name="isXA" type="BOOLEAN">true</property>
                 </graph-query>
             </when>
             <perform>
--- a/rules-reviewed/technology-usage/database-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/database-technology-usage.windup.xml
@@ -1,0 +1,64 @@
+<?xml version="1.0"?>
+<ruleset id="technology-usage-database" xmlns="http://windup.jboss.org/schema/jboss-ruleset" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
+            This ruleset provides statistical summaries of the database items that were found during the analysis.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.7.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.7.0.Final" />
+        </dependencies>
+        <phase>PostMigrationRulesPhase</phase>
+    </metadata>
+    <rules>
+        <rule id="technology-usage-database-01000">
+            <when>
+                <graph-query discriminator="JdbcDatasourceModel"></graph-query>
+            </when>
+            <perform>
+                <technology-identified name="JDBC datasources">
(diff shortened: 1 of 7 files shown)


mrizzi:
retest

rules-reviewed/technology-usage/database-technology-usage.windup.xml
```diff
@@ -0,0 +1,78 @@
...
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,2.7.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,2.7.0.Final" />
+        </dependencies>
+        <phase>PostMigrationRulesPhase</phase>
+    </metadata>
+    <rules>
+        <rule id="technology-usage-database-01000">
+            <when>
+                <graph-query discriminator="JdbcDatasourceModel"  />
```
mrizzi: I haven't been able to find a rule that adds `JdbcDatasourceModel` to the graph so this rule was always failing its test.
I think that it might be better to switch to use `DataSourceModel` instead.

rules-reviewed/technology-usage/database-technology-usage.windup.xml
```diff
@@ -0,0 +1,78 @@
...
+                    <tag name="Store"/>
+                    <tag name="Database"/>
+                    <tag name="Java EE"/>
+                </technology-identified>
+            </perform>
+        </rule>
+        <rule id="technology-usage-database-01001">
+            <when>
+                <graph-query discriminator="DataSourceModel">
+                    <property name="isXA">true</property>
```
mrizzi: This rule is failing its test because this condition is not working properly unless you specify that the property is not a string but a boolean.
You can do it adding the `type` attribute in this way:
```xml
                    <property name="isXA" type="BOOLEAN">true</property>
```
This was an issue raised during #270

rules-reviewed/technology-usage/database-technology-usage.windup.xml
```diff
@@ -0,0 +1,78 @@
...
+                <technology-identified name="JPA named queries">
+                    <tag name="Store"/>
+                    <tag name="Persistence"/>
+                    <tag name="Java EE"/>
+                </technology-identified>
+            </perform>
+        </rule>
+        <rule id="technology-usage-database-01300">
+            <when>
+                <graph-query discriminator="JPAPersistenceUnitModel"  />
```
mrizzi: This rules is failing the test because when it's executed, the method [TechnologyIdentified.perform](https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java#L130) throws the exception `The method org.jboss.windup.graph.model.HasApplications.getApplications has no annotations, therefore frames cannot handle the method.` at [line 130](https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java#L130).
I solved adding, at the end of [JPAPersistenceUnitModel](https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/model/JPAPersistenceUnitModel.java#L83) class, the following code:
```java
    @JavaHandler
    @Override
    Iterable<ProjectModel> getApplications();

    @JavaHandler
    @Override
    boolean belongsToProject(ProjectModel projectModel);

    abstract class Impl implements JPAPersistenceUnitModel, JavaHandlerContext<Vertex>
    {
        public Iterable<ProjectModel> getApplications()
        {
            return this.getJPAConfigurationFileModel().getApplications();
        }

        public boolean belongsToProject(ProjectModel projectModel)
        {
            return this.getJPAConfigurationFileModel().belongsToProject(projectModel);
        }
    }
```

rules-reviewed/technology-usage/tests/database-technology-usage.windup.test.xml
```diff
@@ -0,0 +1,79 @@
...
+                    </not>
+                </when>
+                <perform>
+                    <fail message="JPA named queries Technology Statistic Not Found" />
+                </perform>
+            </rule>
+            <rule id="technology-usage-database-01300-test">
+                <when>
+                    <not>
+                        <technology-statistics-exists name="Persistence units" number-found="1">
```
mrizzi: For this test to work, it's has to be added in the `data/database/` folder at least one file with a persistence unit like [persistence.xml](https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/technology-usage/tests/data/jta/persistence.xml).
With such a file also test `technology-usage-database-01000-test` and `technology-usage-database-01001-test` succeed then.

rules-reviewed/technology-usage/tests/database-technology-usage.windup.test.xml
```diff
@@ -0,0 +1,79 @@
...
+                            <tag name="Database"/>
+                            <tag name="Java EE"/>
+                        </technology-statistics-exists>
+                    </not>
+                </when>
+                <perform>
+                    <fail message="JDBC datasources Technology Statistic Not Found" />
+                </perform>
+            </rule>
+            <rule id="technology-usage-database-01001-test">
```
mrizzi: See comment below for test `technology-usage-database-01300-test`

rules-reviewed/technology-usage/tests/database-technology-usage.windup.test.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" ?>
+<ruletest id="technology-usage-database-test" xmlns="http://windup.jboss.org/schema/jboss-ruleset">
+    <testDataPath>data/database/</testDataPath>
+    <rulePath>../database-technology-usage.windup.xml</rulePath>
+    <ruleset>
+        <rules>
+            <rule id="technology-usage-database-01000-test">
```
mrizzi: See comment below for test `technology-usage-database-01300-test`

rules-reviewed/technology-usage/database-technology-usage.windup.xml
klinki: Thanks Marco, I didn't realize that

rules-reviewed/technology-usage/database-technology-usage.windup.xml
klinki: You are right. I thought there is some relation between `DataSourceModel` and `JdbcDataSourceModel`, but it seems there isn't.

I was quite confused about this too and didn't solve it yet. I guess using just `DataSource` would be better...

rules-reviewed/technology-usage/tests/database-technology-usage.windup.test.xml
klinki: ?  I don't get it? There is data source in `mysql-ds.xml` file. We don't identify these? We identify only data sources in `persistence.xml` ?

rules-reviewed/technology-usage/tests/database-technology-usage.windup.test.xml
klinki: I added `persistence.xml` file. But tests `0100` and `01001` should not be related to that. There is `mysql-ds.xml` file which defines data sources. Or it needs to be in `persistence.xml` file?

rules-reviewed/technology-usage/database-technology-usage.windup.xml
klinki: Nice catch. I'm not sure why I didn't get that exception :( Maybe I missed that...

rules-reviewed/technology-usage/database-technology-usage.windup.xml
mrizzi: yep, i think that using `DataSource` will be good for this rule.
Maybe we can also think about removing `JdbcDatasourceModel` because it's quite useless right now.

rules-reviewed/technology-usage/database-technology-usage.windup.xml
mrizzi: I saw the exception only when i debugged the rule to understand why it was not working as expected.

rules-reviewed/technology-usage/database-technology-usage.windup.xml
klinki: I created JIRA for that. I  will eventually get back to it and remove it

rules-reviewed/technology-usage/tests/database-technology-usage.windup.test.xml
mrizzi: I didn't find any Java rule for creating a `DataSourceModel` from a  `*-ds.xml` file.
`*-ds.xml` files are only handled by rules in [`eap4-xml-config.windup.xml`](https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/eap6/jboss-eap4/eap4-xml-config.windup.xml#L31) file but those rules creates only hints right now.
You can add in  `eap4-eap6-25000`, `eap4-eap6-26000` and `eap4-eap6-27000` rules the tag
```xml
<technology-tag level="INFORMATIONAL">Datasource</technology-tag>
```
and then, in `technology-usage-database-01000` rule, replace the current `<when>` block with:
```xml
            <when>
                <graph-query discriminator="TechnologyTagModel">
                    <property name="name">Datasource</property>
                </graph-query>
            </when>
```

klinki: It seems like we really identify datasources only in `persistence.xml` file by using  `DiscoverJpaConfigurationXmlRuleProvider`.

I would rather create some special rule for identifying datasources, then modifying those you mentioned. Because the won't be executed in case of selecting any other target, then EAP 6, right?

mareknovotny:
retest

mareknovotny:
I wrongly removed the label

mrizzi:
Once windup/windup#1201 will be fine, then i think this can be merged.

mareknovotny:
retest

PR #273, opened 2017-09-25
WINDUPRULE-296: adding javaee rules for Technology stats

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-296

### DIFF
--- a/rules-reviewed/technology-usage/tests/javaee-technology-usage.windup.test.xml
+++ b/rules-reviewed/technology-usage/tests/javaee-technology-usage.windup.test.xml
@@ -1,0 +1,48 @@
+<?xml version="1.0"?>
+<ruletest id="javaee-technology-usage-test" xmlns="http://windup.jboss.org/schema/jboss-ruleset">
+    <testDataPath>data/javaee/</testDataPath>
+    <rulePath>../javaee-technology-usage.windup.xml</rulePath>
+    <ruleset>
+        <rules>
+            <rule id="javaee-technology-usage-00010-test">
+                <when>
+                    <not>
+                        <technology-statistics-exists name="Java EE Batch" number-found="1">
+                            <tag name="Batch"/>
+                            <tag name="Java EE"/>
+                        </technology-statistics-exists>
+                    </not>
+                </when>
+                <perform>
+                    <fail message="'Java EE Batch' Technology Statistic Not Found" />
+                </perform>
+            </rule>
+            <rule id="javaee-technology-usage-00020-test">
+                <when>
+                    <not>
+                        <technology-statistics-exists name="Java EE CDI" number-found="1">
+                            <tag name="CDI"/>
+                            <tag name="Java EE"/>
+                        </technology-statistics-exists>
+                    </not>
+                </when>
+                <perform>
+                    <fail message="'CDI Bean' Technology Statistic Not Found" />
+                </perform>
+            </rule>
+            <rule id="javaee-technology-usage-00030-test">
+                <when>
+                    <not>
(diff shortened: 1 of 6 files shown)


rules-reviewed/technology-usage/tests/javaee-technology-usage.windup.test.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0"?>
+<ruletest id="javaee-technology-usage-test" xmlns="http://windup.jboss.org/schema/jboss-ruleset">
+    <testDataPath>data/javaee/</testDataPath>
+    <rulePath>../javaee-technology-usage.windup.xml</rulePath>
+    <ruleset>
+        <rules>
+            <rule id="javaee-technology-usage-00010-test">
+                <when>
+                    <not>
+                        <technology-statistics-exists name="Java EE Batch" number-found="1">
```
mareknovotny: This is a little confusing to me that number-found is only 1 instead of 2

jsight: It is because only a single TechnologyTagModel is created (although the model itself links to multiple files).

jsight: @mrizzi - I don't think that it creates two of them. They are shared as long as the technology tag name is the same.

mareknovotny: I understand that from https://github.com/windup/windup-rulesets/pull/273#discussion_r140862779 but I was more thinking how it should work to get the technology stats counting properly, I can be totaly wrong from some my presumption. Is our goal to have aggregated technology stats for an application?

mrizzi: @jsight +1 you're absolutely right!
[TechnologyTagService.addTagToFileModel](https://github.com/windup/windup/blob/master/reporting/api/src/main/java/org/jboss/windup/reporting/service/TechnologyTagService.java#L50) adds just one `TechnologyTagModel`

jsight: @mareknovotny - If an accurate count is necessary, then it probably can't use the technology tag. I think some of these are just on/off in the report, though, rather than needing a detailed count.

mareknovotny: @jsight right, true/false on some technology makes sense. The only issue with technology-tag now after I have been using it is that the implication for that is hidden and that could confuse the possible authors too not only us. I didn't realize while reviewing the implementation PR, but using only xml rules is easy even for  counting too or not?

jsight: I agree... it is sort of hidden. It would be nice if we can find a better solution for that, as it is likely to cause an issue at some point.

What do you propose doing?

mareknovotny: just use number of linked FileModels as a statistic counter?

jsight: That is possible. It would also have to group them by project, though. It would make sense to do that for Classification as well. Do you want to file a JIRA with this?

mareknovotny: @jsight I created https://issues.jboss.org/browse/WINDUP-1703

jsight: Great... thanks!

PR #280, opened 2017-10-04
WINDUPRULE-296: fixed java ee technology tags

mareknovotny:
fixing WINDUPRULE-296 after @OndraZizka catch it doesn't have what the doc specified.

### DIFF
--- a/rules-reviewed/technology-usage/javaee-technology-usage.windup.xml
+++ b/rules-reviewed/technology-usage/javaee-technology-usage.windup.xml
@@ -31,19 +31,19 @@
                 </xmlfile>
             </when>
             <perform>
-                <technology-tag level="INFORMATIONAL">Java EE Batch API</technology-tag>
+                <technology-tag level="INFORMATIONAL">Java EE Batch</technology-tag>
             </perform>
         </rule>
-        <!-- Creating technology stats for Java EE Batch -->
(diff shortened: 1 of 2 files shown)


OndraZizka:
I'll leave up to Jess to confirm if this is how it should be. It matches how I understood it should be.

rules-reviewed/technology-usage/javaee-technology-usage.windup.xml
```diff
@@ -58,41 +58,41 @@
...
         </rule>
-        <!-- JSON-P identication -->
+        <!-- JSON-P identification -->
         <rule id="javaee-technology-usage-00030">
             <when>
                 <javaclass references="javax.json.{*}"/>
             </when>
             <perform>
-                <technology-tag level="INFORMATIONAL">Java EE JSON-P</technology-tag>
+                <technology-tag level="INFORMATIONAL">JSON-P</technology-tag>
```
jsight: I think it is important to keep the "Java EE" prefix (or something like it). "JSON-P" is an overloaded acronym, and the more common usage isn't the one referenced here.

jsight:
I just had one comment.... +1 to everything else.

mareknovotny:
reverted back Java EE also for Batch just to be sure it is clear

PR #329, opened 2018-05-21
WINDUPRULE-344: Started on test generator xslt

jsight:


jsight:
Merging as this doesn't affect the build output and is already useful...

PR #342, opened 2018-09-14
Oracle2openjdk

jeichler:
discard https://github.com/windup/windup-rulesets/pull/339 and use this instead


windupgithubbot1:

PR #351, opened 2018-11-08
WINDUPRULE-178: add text to hint to aid user choosing the right value…

m-brophy:
… for the new hibernate.transaction.coordinator_class property:

### DIFF
--- a/rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
+++ b/rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
@@ -30,7 +30,8 @@
                         Next the contract in `hibernate.transaction.coordinator_class` property should refer to `org.hibernate.resource.transaction.TransactionCoordinatorBuilder`
                         instead of to `org.hibernate.engine.transaction.spi.TransactionFactory`
 
-                        If a JPA application does not provide a setting for hibernate.transaction.coordinator_class, Hibernate will automatically build the proper transaction coordinator based on the transaction type for the persistence unit.
+                        If a JPA application does not provide a setting for `hibernate.transaction.coordinator_class`, Hibernate will automatically build the proper transaction coordinator based on the transaction type for the persistence unit.
+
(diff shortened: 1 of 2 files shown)


rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -29,6 +29,9 @@
 
                         Next the contract in `hibernate.transaction.coordinator_class` property should refer to `org.hibernate.resource.transaction.TransactionCoordinatorBuilder`
                         instead of to `org.hibernate.engine.transaction.spi.TransactionFactory`
+
+                        If a JPA application does not provide a setting for hibernate.transaction.coordinator_class, Hibernate will automatically build the proper transaction coordinator based on the transaction type for the persistence unit.
```
mrizzi:
```suggestion
                        If a JPA application does not provide a setting for `hibernate.transaction.coordinator_class`, Hibernate will automatically build the proper transaction coordinator based on the transaction type for the persistence unit.

```

rules-reviewed/eap7/eap6/hibernate4-xml.windup.xml
```diff
@@ -29,6 +29,9 @@
...
+                        If a non-JPA application does not provide a setting for hibernate.transaction.coordinator_class, Hibernate will use jdbc as the default. This default will cause problems if the application actually uses JTA-based transactions. A non-JPA application that uses JTA-based transactions should explicitly set hibernate.transaction.coordinator_class=jta or provide a custom org.hibernate.resource.transaction.TransactionCoordinatorBuilder that builds a org.hibernate.resource.transaction.TransactionCoordinator that properly coordinates with JTA-based transactions.
```
mrizzi:
```suggestion
                        If a non-JPA application does not provide a setting for `hibernate.transaction.coordinator_class`, Hibernate will use jdbc as the default. This default will cause problems if the application actually uses JTA-based transactions. A non-JPA application that uses JTA-based transactions should explicitly set `hibernate.transaction.coordinator_class=jta` or provide a custom `org.hibernate.resource.transaction.TransactionCoordinatorBuilder` that builds a `org.hibernate.resource.transaction.TransactionCoordinator` that properly coordinates with JTA-based transactions.
```

PR #354, opened 2019-02-14
WINDUPRULE-375: change rule to fire when target is eap:7

m-brophy:


PhilipCattanach:
The testing was successful please merge this change into the Master branch.

PR #357, opened 2019-02-26
WINDUPRULE-377: hibernate 5.1-5.3 migration, hibernate-java8 in pom.xml

m-brophy:
dependent on https://github.com/windup/windup/pull/1348

This PR includes solutions for both WINDUPRULE-377 and WINDUPRULE-378 - they are very similar and can both be solved in the same rule by parameterizing it and passing in the artifactIds we wish to search for in each case: 

hibernate-java8 for WINDUPRULE-377, 
hibernate-entitymanager for WINDUPRULE-378

### DIFF
--- a/rules-reviewed/eap7/eap71/hibernate51-53.windup.xml
+++ b/rules-reviewed/eap7/eap71/hibernate51-53.windup.xml
@@ -24,17 +24,30 @@
             </perform>
         </rule>
         -->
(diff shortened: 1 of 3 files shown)


m-brophy:
This test is failing because the change for parameterization in artifactId has to be committed in windup/windup here #https://github.com/windup/windup/pull/1348 for this one to work

mrizzi:
retest

PhilipCattanach:
Tested successfully

PR #368, opened 2019-03-07
Winduprule 385 hibernate limit offset

jonathanvila:
https://issues.jboss.org/browse/WINDUPRULE-385


PhilipCattanach:
Jonathan, 
I have tried running this locally and the test failed.
I suspect you've built it and tested it in isolation, and another Hibernate test file has triggered the rule
https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/eap7/eap6/tests/data/data-hibernate4/hibernate.cfg.xml#L9
So the expected count and actual count of incidents is different 
Also I think the file that has triggered the rule, should not have done
<property name="hibernate.dialect">org.hibernate.dialect.Oracle12cDialect</property>
but Oracle12cDialect is not one of the required dialects

jonathanvila:
@PhilipCattanach the issue is that I didnt put a WIP in the PR as I still had to test after moving the rule to hibernate4 file.
Oracle12 Dialect , you are right. As there is also a test for the no-namespace property this can lead to an error.
I'm checking this currently.

PhilipCattanach:
Tested the PR successfully  - All good

PR #371, opened 2019-03-22
WINDUP-2337: tech report jax-rs references in rest box rather than http

m-brophy:
dependent on https://github.com/windup/windup/pull/1350


m-brophy:
@mrizzi see  https://github.com/windup/windup/pull/1350, required to make this work

PR #390, opened 2019-12-17
Camel2 camel3

johnpoth:
The Last commit contains the problematic substitution, thanks !


windupgithubbot1:

mrizzi:
it seems like you found a bug @johnpoth so i filled https://issues.redhat.com/browse/WINDUP-2525.
I implemented some changes to, at least, let you work.
Instruction to use them:
```
$ git clone https://github.com/mrizzi/windup.git
$ cd windup/
$ git checkout --track origin/WINDUP-2525
$ mvn clean install -DskipTests
```
then in your `windup-ruleset` project change the in https://github.com/windup/windup-rulesets/blob/master/pom.xml#L26 the `version.windup` to be `4.3.1-SNAPSHOT` (instead of `4.3.0.Final`) so that you will use the artifacts you will have just built.
Then the rules in this PR should work

mrizzi:
Talked with @johnpoth about this PR and it has been used to share the code and implement the fix in https://github.com/windup/windup/pull/1372 but this won't be merged so i'm going to close it.

PR #407, opened 2020-02-07
WINDUPRULE-458 camel-kafka circularTopicDetection and bridgeEndpoint …

johnpoth:
…endpoint options have been removed

Hi @mrizzi  I've noticed that if I use an [or](https://github.com/windup/windup-rulesets/compare/master...johnpoth:WINDUPRULE-458?expand=1#diff-089aca84fa17d063e666047e9e7c2ffcR18) in my rule, each match within the `or` will only be triggered once per test file so my test is failing due to the number of errors reported being lower than the actual number of errors present in the files.

I noticed that if I create seperate rules instead of using `or`, the number of errors is consistent with the number of errors present.

I this a known issue ? Thanks again !


windupgithubbot1:

johnpoth:
Hi @m-brophy @mrizzi, have you had a chance to look at this ? It might be that it's expected to only report the first error and the user is expected to run windup once more after correcting the first bug ? If so then I believe the PR is ready to be reviewed and merged. Thanks !

m-brophy:
Hi @johnpoth I've discussed this with @mrizzi and we think this is how the rules are supposed to work - if anything within an or fires, the perform section executes but not multipletimes corresponding to the number of conditions met in the or. If you want each of these condition instances to fire then they need to be in separate rules.

johnpoth:
I've separated the test cases so all the rules are fired. If this becomes a problem we'll have to migrate to groovy to avoid too much duplication.

Ready to be reviewed. Thanks!

PR #424, opened 2020-03-30
[WINDUPRULE-479] maven-javax-to-jakarta links update

emmartins:
Issue: https://issues.redhat.com/browse/WINDUPRULE-479


emmartins:
@mrizzi done

PR #469, opened 2020-07-23
Winduprule 615 spring security

jonathanvila:
https://issues.redhat.com/browse/WINDUPRULE-615


rules-reviewed/quarkus/springboot/springboot-security-to-quarkus.windup.xml
```diff
@@ -0,0 +1,34 @@
...
+            <addon id="org.jboss.windup.rules,windup-rules-xml,3.0.0.Final" />
+        </dependencies>
+        <sourceTechnology id="springboot" />
+        <targetTechnology id="quarkus" />
+    </metadata>
+    <rules>
+        <rule id="springboot-security-to-quarkus-00000">
+            <when>
+                <dependency groupId="org.springframework.security" artifactId="spring-security-core"/>
+            </when>
```
m-brophy, line 20:
```suggestion
            <when>
            <project>
                <artifact groupId="org.springframework.security" artifactId="spring-security-core"/>
                </project>
            </when>
```
If you build a test data jar from spring initializr with the security dependency it will do it using a dependency of the security starter artifact in the pom. It will build with the starter-core jar we're looking for here in the lib folder and this rule won't find that. You'd need a rule like the one I suggest here. This works fine but I don't think it's as realistic a test scenario as the one I've outlined.

rules-reviewed/quarkus/springboot/springboot-security-to-quarkus.windup.xml
```diff
@@ -0,0 +1,34 @@
...
+            <perform>
+                <hint title="Replace the SpringBoot Security artifact with Quarkus 'spring-security' extension" effort="1" category-id="mandatory">
+                    <message>
+                        Replace the SpringBoot Security artifact with Quarkus `spring-security` extension
+
+                        Spring Security is in spring-security-core artifact brought transitively by `org.springframework.boot:spring-boot-starter-security` dependency.  
+                        Add Quarkus dependency `io.quarkus:quarkus-spring-security`
+                    </message>
+                    <link title="Quarkus DI Guide" href="https://quarkus.io/guides/spring-security" />
```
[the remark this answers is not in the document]
jonathanvila: Good catch !!! :+1:

rules-reviewed/quarkus/springboot/springboot-security-to-quarkus.windup.xml
jonathanvila: I'm not sure I follow you.
I've created a project with start.spring.io , including Security.
When you build the project , and later you unzip the jar, you can see that in the lib folder there's an artifact "spring-security-core", that it's the one we are checking.

m-brophy: I've checked that an analysis finds this file in the lib folder of a jar which was the scenario I was concerned about, so I'm happy this point doesn't need to be addressed any further.

m-brophy (approved):

jonathanvila:
retest

jonathanvila:
Merging on Marks behalf.

PR #471, opened 2020-07-28
refactor: disable sourceMode flag not needed

aureamunoz:
@jonathanvila As we've seen I've disabled this flag when not needed


windupgithubbot1:

PR #490, opened 2020-09-18
WINDUPRULE-627: spring flo to quarkus rule

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-627

### DIFF
--- a/rules-reviewed/quarkus/springboot/springboot-flo-to-quarkus.windup.xml
+++ b/rules-reviewed/quarkus/springboot/springboot-flo-to-quarkus.windup.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0"?>
+<ruleset xmlns="http://windup.jboss.org/schema/jboss-ruleset" id="springboot-flo-to-quarkus"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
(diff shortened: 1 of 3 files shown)


rules-reviewed/quarkus/springboot/springboot-flo-to-quarkus.windup.xml
```diff
@@ -0,0 +1,32 @@
...
+    </metadata>
+    <rules>
+        <rule id="springboot-flo-to-quarkus-00000">
+            <when>
+                <filecontent pattern="flo-editor"></filecontent>
+            </when>
+            <perform>
+                <hint title="Spring Flo is not supported by Quarkus" effort="13" category-id="mandatory">
+                    <message>
+                        Spring Flo is not supported by Quarkus
```
jonathanvila, line 24: Apparently Spring Flo is a Javascript library.... so, I don't see any relation with Quarkus. It could be packaged with a Quarkus application as this will run on the browser side as any other Javascript library.

Wdyt ?

m-brophy: I agree that running the javascript will be fine. I think maybe the reason that it's listed as not supported is that it's used to produce the visual stream builder for Spring Cloud Data Flow which definitely isn't supported. So maybe there's a case to change the message to something like:

 "Spring Flo Javascript library has been detected. This is used in conjunction with Spring Cloud Data Flow which is not supported in Quarkus"

wdyt?

jonathanvila: I agree :+1:

jonathanvila (approved):

jonathanvila (approved):
LGTM :+1:

PR #520, opened 2021-01-06
WINDUPRULE-702: prevent noisy rule triggering

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-702


PhilipCattanach:
Tested successfully

PR #527, opened 2021-02-04
WINDUPRULE-711 Quarkus Release 1.11 - MongoDB Panache

PhilipCattanach:
https://issues.redhat.com/browse/WINDUPRULE-711

### DIFF
--- a/rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
+++ b/rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0"?>
+<ruleset xmlns="http://windup.jboss.org/schema/jboss-ruleset" id="quarkus1-11-mongodb-panache"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
+            This ruleset gives hints for upgrading to quarkus 1.11 for applications that use MongoDB Panache
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,3.0.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,3.0.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-xml,3.0.0.Final" />
+        </dependencies>
+        <!-- The major version is implied by the target, the minor is reflected in the verionRange -->
+        <sourceTechnology id="quarkus1" versionRange="(,10]"/>
+        <targetTechnology id="quarkus1" versionRange="[11,)"/>
+    </metadata>
+    <rules>
+        <rule id="quarkus1-11-mongodb-panache-00000">
+            <when>
+                <graph-query discriminator="WindupJavaConfigurationModel" as="discard">
+                    <property type="BOOLEAN" name="sourceMode">true</property>
+                </graph-query>
+                <project>
+                    <artifact groupId="io.quarkus" artifactId="quarkus-mongodb-panache"/>
+                </project>
+            </when>
+            <perform>
+                <iteration over="default">
+                    <hint title="MongoDB Panache changed to conform with Hibernate Panache behavior" effort="1" category-id="potential">
+                        <message>
+                            A recent change was made to MongoDB panache to bring it in to conformity with the Hibernate Panache behavior. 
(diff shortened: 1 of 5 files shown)


rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
```diff
@@ -0,0 +1,69 @@
...
+            This ruleset gives hints for upgrading to quarkus 1.11 for applications that use MongoDB Panache
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,3.0.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,3.0.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-xml,3.0.0.Final" />
+        </dependencies>
+        <!-- The major version is implied by the target, the minor is reflected in the verionRange -->
+        <sourceTechnology id="quarkus1" versionRange="(,10]"/>
+        <targetTechnology id="quarkus1" versionRange="[11,)"/>
```
jonathanvila, line 15: Is the intention of this ruleset to cover migration **from** any Quarkus version **older than 11** to any version **newer or equal to 11**, so this rule will be executed for any target 11, 12, 13, 14, etc... ?

rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
```diff
@@ -0,0 +1,69 @@
...
+    </metadata>
+    <rules>
+        <rule id="quarkus1-11-mongodb-panache-00000">
+            <when>
+                <graph-query discriminator="WindupJavaConfigurationModel" as="discard">
```
jonathanvila, line 20: why constraining to a source mode analysis ?

rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
```diff
@@ -0,0 +1,69 @@
...
+                </project>
+            </when>
+            <perform>
+                <iteration over="default">
+                    <hint title="MongoDB Panache changed to conform with Hibernate Panache behavior" effort="1" category-id="potential">
+                        <message>
+                            A recent change was made to MongoDB panache to bring it in to conformity with the Hibernate Panache behavior. 
+                            Public field accesses on `MongoEntity` and `ReactiveMongoEntity` types are now wrapped with the appropriate `get` or `set` methods. 
+                            In general, you will like not see any difference in your application. 
+                            However, if you have written a custom `get` or `set` method you may notice a change in behavior if those custom methods deviate from the standard `get`/`set` paradigm.
```
jonathanvila, line 34: seems this is a general advise but it can lead to no changes  if the user has not written any custom get or set method...... shouldnt we check first if the user is really doing that action ?

rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
```diff
@@ -0,0 +1,69 @@
...
+        </rule>
+        <rule id="quarkus1-11-mongodb-panache-00010">
+            <when>
+                <or>
+                    <javaclass references="io.quarkus.mongodb.panache.PanacheMongoEntity">
+                        <location>INHERITANCE</location>
+                    </javaclass>
+                    <javaclass references="io.quarkus.mongodb.panache.PanacheMongoEntityBase">
+                        <location>INHERITANCE</location>
+                    </javaclass>
```
jonathanvila, line 50: if user doesnt write its own getter/setter , then will be no problem, according to doc, so why raising this issue always.

rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
PhilipCattanach: That's correct so you could be migrating from say 1.8 to 1.13. so you would want the rules that cover the 1.10  to 1.11 to fire.

rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
PhilipCattanach: It's just an awareness situation. We need to be proportionate with these rules. So I agree we could be super precise. But I don't think that is warranted. Just alert he user to the possibility and let them make the decision as to whether to drill into it further.

rules-reviewed/quarkus1/quarkus1-10/quarkus1-11-mongodb-panache.windup.xml
PhilipCattanach: As above Jonathan - it's an awareness thing and getting the balance right in terms of practicality.

PR #531, opened 2021-02-24
WINDUPRULE-715 Quarkus Release 1.12 - Fast-jar

PhilipCattanach:


m-brophy (commented):
Should these files not be in a quarkus1-12 folder rather than the existing quarkus1-11 one?

PR #534, opened 2021-02-28
WINDUPRULE-721 Quarkus Release 1.12 - Quarkus Maven Plugin

PhilipCattanach:
https://issues.redhat.com/browse/WINDUPRULE-721

This is a groovy rule that identifies the presence of the Quarkus maven plugin via FileContent.matches("quarkus-maven-plugin").inFileNamed("pom.xml")
I am not altogether happy with it as the pom.xml file is lot listed on the right hand side if the issues report and I can't get the message body to format the sample xml properly. Tabs and space characters appear to be ignored so the example XML is not indented nicely.


mrizzi (commented):
Suggested some changes in https://github.com/PhilipCattanach/windup-rulesets/pull/2

mrizzi (approved):
Thanks @PhilipCattanach :+1:

PR #537, opened 2021-03-01
WINDUPRULE-722: quarkus1-12 distroless image rule

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-722


PhilipCattanach:
Hi @m-brophy just a couple of minor things
1). The description tag is still referring to Fast-jar as default rather than distroless image
2). The message needs to change the background color of the image names. So encapsulate the image names with ` `

PhilipCattanach:
All good - thanks @m-brophy

PR #546, opened 2021-03-24
WINDUPRULE-734 - Quarkus 1.13 maven uber-jar

PhilipCattanach:


m-brophy:
I've created a PR to this PR to fix the build error Phil.
native-image-goal test is picking up other pom files and that makes the count wrong so I've changed the testdatapath in that file to point to the particular pom we're interested in.

PR #556, opened 2021-05-04
Winduprule 743

PhilipCattanach:
Added thorntail to all of the rules as a sourceTechnology.
I have built the cli and executed al of the thorntail rules with thorntail specified as the source. All the rules fired.


mrizzi (approved):
LGTM, thanks @PhilipCattanach :+1:

PR #565, opened 2021-10-14
Windup 3148

PhilipCattanach:
Notes about testing these rules.
The rules cover the artifacts highlighted in green on [this|https://docs.google.com/spreadsheets/d/1EfsUAlZpciZPF547lFwRuph0uh61na-0CYeAAtGpKMI/edit#gid=445606470] worksheet.
To test the quickfixes use the MTA Code Ready Studio plugin.
Copy the MTA CLI distribution of that contains the rules. It can be found [here|https://drive.google.com/file/d/1aR0rV-TnkfrVAUGwvCsM1AvJUKFBJ1dX/view?usp=sharing]. 
Use the cog icon within the MTA perspective in CRS to point to that MTA CLI (so the rules are included).
Take a copy of the pom.xml file used to drive the rules tests /windup-rulesets/rules-reviewed/eap8/eap7/tests/data/dependencies/pom.xml
Open that file in CRS and run an MTA analysis using the advanced option target jakarta-ee
Open the pom file and ensure all of the quickfixes work.

@m-brophy - As well as testing the PR would you mind reviewing the Googlesheet that maps the javaee artifacts to the jakarta equivalents. It would be comforting to know I got the mapping right.

PR #569, opened 2021-10-19
Winduprule 767

PhilipCattanach:
This ruleset was using the project tag to identify artifacts that had to be renamed. 
But a quickfix based on the project can only have 1 find and replace.
If both the groupId and artifactId had to be replaced then the separate rules were required for each.
As a consequence this ruleset has been refactored.

The rules groupId's and artifactId's that are subject to the refactored rules have 'Yes' with a green background colour in column F of this Googlesheet.  https://docs.google.com/spreadsheets/d/1EfsUAlZpciZPF547lFwRuph0uh61na-0CYeAAtGpKMI/edit#gid=445606470

I would suggest testing this PR before looking at related PR - https://github.com/windup/windup-rulesets/pull/565

A snapshot build of the CLI containing this PR can be found here: https://drive.google.com/drive/folders/1XqjQkiH3JnL5qOPoHvvTyxmPuRpd-YQd

Point CRS to use that distribution of the CLI and analyze a copy of the pom.xml provided in the data/data-maven-javax-to-jakarta
to ensure the quickfixes work.


PhilipCattanach:
@m-brophy I ave reverted back the when criteria for the g:javax.activation in this ruleset to use the project construct so it only will fire if the g:javax.activation AND a:activation
We have to accept that the quick-fix might not work in all IDEs but hopefully that is something John will be able to sort out.
A little bit of a compromise I know but I think it is a practical one.

PR #582, opened 2021-11-30
WINDUPRULE-787: remove eap8 target for rules

m-brophy:


PhilipCattanach:
@m-brophy - Thanks Mark - The changes look fine. The rules are run fine when tested independently and when an application is analyzed with the jakarta-ee target from the CLI. No mention of the eap:8 target when using the --listTargetTechnologies argument. Just waiting for the automated checks to complete then I will merge.

PR #599, opened 2022-01-19
WINDUPRULE-807 - Tech Tags - Spring Integration

PhilipCattanach:
Another tech tags change. This one is no reliant on any of the new tiers.


PhilipCattanach:
@m-brophy - Hi Mark - good spot. I have updated the PR can you please re-test.

PR #603, opened 2022-01-25
Winduprule 812

PhilipCattanach:
This PR adds some additional rules to the technology-usage/embedded-framework* rulesets.
While adding additional rules for the Spring Shell technology I also spotted that the sequence of the rule-id's were wrong so they had to be renumbered. Also there is a catch-all rule that needs to be the the final one in the embedded-framework.windup.xml ruleset. So that catch-all rule was moved to the bottom of the ruleset. 
In order to test the Spring Shell rule you will have to build locally a version of the CLI containing https://github.com/windup/windup/pull/1443


carlosthe19916 (approved):
LGTM

PR #615, opened 2022-02-09
[WINDUPRULE-816] Add rule for detecting quarkus-spring-web to add the jackson reactive dependency

jmle:
See [WINDUPRULE-816](https://issues.redhat.com/browse/WINDUPRULE-816).
I have unified the pom-fetching rule and the artifact-fetching one into a single rule with an `or` clause. It could be split if needed.


PhilipCattanach:
Thanks @jmle 
The rule looks good and I have tested successfully. I see in your latest commit you have updated the message to

"Starting with Quarkus version 2.5, the underlying JAX-RS engine must be chosen. For performance reasons, the reactive-jackson is selected by default."

The way I interpreted the guide is that the user must make the choice and update the pom file. So perhaps the wording below might be better? 

"Starting with Quarkus version 2.5, the underlying JAX-RS engine must be chosen. For performance reasons the `quarkus-resteasy-reactive-jackson` dependency is recommended."

jmle:
Hi @PhilipCattanach, the problem with stating that the rule is _recommended_ is that it might lead the user to believe that another option is possible. In this case, if the user chooses the other alternative (the classic resteasy jackson dependency) the rule would still fail... What do you think?

PhilipCattanach:
@jmle good point. 
So I would suggest we replace 'is recommended' with 'is strongly recommended' or with 'should be used'.
So if they choose to use the alternative dependency MTA is still encouraging them to make the _right choice_.

jmle:
@PhilipCattanach thanks, I just reworded the sentence

PhilipCattanach:
@jmle - Great. Retested successfully. I will merge as soon as the automated tests complete.

PR #616, opened 2022-02-09
[WINDUPRULE-817] Add rule to detect @SpringBootApplication annotation

jmle:
See [WINDUPRULE-817](https://issues.redhat.com/browse/WINDUPRULE-817)


PhilipCattanach:
@jmle Thanks for the PR. I have a few minor comments........

I think the rule message could be expanded to add a bit more detail about option 2.

Add the org.springframework.boot:spring-boot-autoconfigure dependency as an optional Maven dependency. An optional dependency is available when an application compiles but is not packaged with the application at runtime. Doing this would allow the application to compile without modification, but you would also need to maintain a Spring version along with the Quarkus application.

There is a commented out link in the rule. That should be removed (or replaced if there is a relevant link in the Quarkus guides).
It would be better if the Fail message contained the id of the Rule so replace
[springboot-annotations-to-quarkus-extension]
with 
[springboot-annotations-to-quarkus-00000-test]

jmle:
Thanks @PhilipCattanach, I have removed the link in this case, since I haven't found any relevant information regarding this particular change.

PhilipCattanach:
@jmle - something I should have picked up on the first review. Can you change word 'adding' to 'add'.
 It just reads better. Then I'll get this merged to master. 
Thanks Juama!

jmle:
@PhilipCattanach done, thanks!

PhilipCattanach:
@jmle - Super! All good. I will merge the PR as soon as the automated tests complete.
Thank you.

PR #621, opened 2022-02-14
WINDUPRULE-809 - Tech tags - Spring Properties

PhilipCattanach:
New rules to identify Spring Properties


rules-reviewed/technology-usage/configuration-management.windup.xml
```diff
@@ -27,5 +27,69 @@
...
+                </classification>
+                <technology-tag level="INFORMATIONAL">Spring Properties</technology-tag>
+            </perform>
+            <where param="profile">
+                <matches pattern=".+" />
+            </where>
+            <where param="extension">
+                <matches pattern="(properties|yml|yaml)" />
+            </where>
+        </rule>
```
m-brophy, line 46: This rule is to find files named in the format 'application-xxxx.properties' (or .yml or .yaml). Do we want it to also find files named just 'application.properties' , as in the rule below it? If so, the profile pattern needs to be the same as that in rule ..-0300

rules-reviewed/technology-usage/tests/data/configuration-management/source0/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.integration</groupId>
+            <artifactId>spring-integration-core</artifactId>
+            <version>1.5.22.RELEASE</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-integration</artifactId>
+            <version>2.3.4.RELEASE</version>
```
m-brophy, line 38: Do we need this changed test file? It doesn't seem to fire any of the new rules.

Same question for the ...RELEASE.jar test file

rules-reviewed/technology-usage/configuration-management.windup.xml
```diff
@@ -27,5 +27,69 @@
...
+                <file filename="application{profile}.{extension}"   />
+            </when>
+            <perform>
+                <classification title="Spring application properties detected" category-id="information" effort="0">
+                    <description>Spring application properties detected</description>
+                </classification>
+                <technology-tag level="INFORMATIONAL">Spring Properties</technology-tag>
+            </perform>
+            <where param="profile">
+                <matches pattern=".+" />
```
m-brophy, line 41: The dot-plus means 'any char one or more times' which means there has to be something there in the file name. If you change it to `<matches pattern=".*"/>` it means zero or more times so it accepts nothing between `application` and `.properties`
```suggestion
                <matches pattern=".*" />
```

m-brophy (changes requested):

m-brophy (approved):

PR #634, opened 2022-03-15
WINDUPRULE-826 SB2Quarkus Swagger and OpenAPI property settings

PhilipCattanach:
This PR contains rules for WINDUPRULE-826 and WINDUPRULE-827
They are both very similar in that they are SB2Quarkus rules pertaining to properties for Swagger and Open API respectively.
These rules originate from this tutorial https://github.com/RedHat-Middleware-Workshops/spring-to-quarkus-todo#migrate-data-source-properties

### DIFF
--- a/rules-reviewed/quarkus/springboot/springboot-properties-to-quarkus.windup.xml
+++ b/rules-reviewed/quarkus/springboot/springboot-properties-to-quarkus.windup.xml
@@ -114,5 +114,55 @@
                 <matches pattern="(properties|yml|yaml)" />
             </where>
         </rule>
+        <rule id="springboot-properties-to-quarkus-00005">
+            <when>
+                <filecontent filename="application{profile}.{extension}" pattern="springdoc.swagger-ui.path" />
+            </when>
+            <perform>
+                <hint title="Replace Spring Swagger endpoint mapping" effort="1" category-id="mandatory">
+                    <message>
+                        Replace `springdoc.swagger-ui.path` with `quarkus.swagger-ui.path`
+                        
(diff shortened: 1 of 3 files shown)


rules-reviewed/quarkus/springboot/springboot-properties-to-quarkus.windup.xml
```diff
@@ -114,5 +114,55 @@
...
+            </where>
+        </rule>
+        <rule id="springboot-properties-to-quarkus-00006">
+            <when>
+                <filecontent filename="application{profile}.{extension}" pattern="springdoc.api-docs.path" />
+            </when>
+            <perform>
+                <hint title="Replace Spring OpenAPI endpoint mapping" effort="1" category-id="mandatory">
+                    <message>
+                        Replace `springdoc.api-docs.path` with `quarkus.smallrye-openapi.path`
```
m-brophy, line 150: do we need to add something here to mention changing `=/openapi` to `=/swagger-ui`?

PhilipCattanach: Hi @m-brophy , 
in both cases the value on the right side of the is consistent, after moving to quarkus.
It is just the name of the property that needs to be changed.

This is the solution view of the file https://github.com/RedHat-Middleware-Workshops/spring-to-quarkus-todo/blob/solution/src/main/resources/application.properties

This is the original SB version of the file
https://github.com/RedHat-Middleware-Workshops/spring-to-quarkus-todo/blob/main/src/main/resources/application.properties

m-brophy (approved):

PR #635, opened 2022-03-18
WINDUPRULE:796 new rule to handle removal of javax.security.auth.Policy

m-brophy:
JDK8 class removed in JDK11

https://issues.redhat.com/browse/WINDUPRULE-796


PhilipCattanach:
Hi @m-brophy The iterable-filter is commented out in the test. It should be active and set to 2.
I'm not sure that we need the link to the Java 10 documentation.  I guess it doesn't hurt.

PhilipCattanach:
Hi @m-brophy The iterable-filter is commented out in the test. It should be active and set to 2.
I'm not sure that we need the link to the Java 10 documentation.  I guess it doesn't hurt.

m-brophy:
Hi @PhilipCattanach , the java 10 docs are in there because they detail the changes needed to use the new class instead of the old one. The Java 11 docs have no mention of the old class (understandably as it's gone) and are there as a reference to the new class. 

I'll reinstate the iterable-filter.

m-brophy:
iterable-filter change made as requested

PR #637, opened 2022-03-21
[WINDUPRULE-785] PropertyChangeListeners removed

jmle:
See [WINDUPRULE-785](https://issues.redhat.com/browse/WINDUPRULE-785).

PR #640, opened 2022-03-22
[WINDUPRULE-791] Add rule for detecting usage of sun.misc base64 encoders

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-791).

### DIFF
--- a/rules-reviewed/openjdk11/openjdk8/java-removals.windup.xml
+++ b/rules-reviewed/openjdk11/openjdk8/java-removals.windup.xml
@@ -152,5 +152,23 @@
                 <matches pattern="(getInt|putInt|getFloat|putFloat|getDouble|putDouble|getBoolean|putBoolean|getObject|putObject|getByte|putByte|getChar|putChar|getShort|putShort|getLong|putLong|fieldOffset|staticFieldBase|tryMonitorEnter|monitorEnter|monitorExit)"/>
             </where>
         </rule>
+        <rule id="java-removals-00101">
+            <when>
+                <or>
+                    <javaclass references="sun.misc.BASE64Encoder"/>
+                    <javaclass references="sun.misc.BASE64Decoder"/>
+                </or>
+            </when>
+            <perform>
+                <hint title="The `sun.misc.Base64Encoder/Decoder` classes have been removed" effort="1" category-id="mandatory">
+                    <message>
+                        The `sun.misc.Base64Encoder/Decoder` classes have been removed. Replace with `java.util.Base64.getMimeEncoder`
+                        and `java.util.Base64.getMimeDecoder` respectively.
+                    </message>
+                    <link title="Code example" href="https://hg.openjdk.java.net/jdk8/jdk8/jdk/rev/9078c34437ab"/>
+                    <link title="Original enhancement request" href="https://bugs.openjdk.java.net/browse/JDK-8006182"/>
+                </hint>
+            </perform>
+        </rule>
     </rules>
 </ruleset>
--- a/rules-reviewed/openjdk11/openjdk8/java-removals.windup.xml
+++ b/rules-reviewed/openjdk11/openjdk8/java-removals.windup.xml
@@ -154,21 +154,20 @@
         </rule>
         <rule id="java-removals-00101">
             <when>
-                <or>
-                    <javaclass references="sun.misc.BASE64Encoder"/>
-                    <javaclass references="sun.misc.BASE64Decoder"/>
-                </or>
+                <javaclass references="sun.misc.BASE64{encoder-decoder}{*}"/>
             </when>
             <perform>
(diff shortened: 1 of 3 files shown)


gnu-andrew:
This looks right to me. I was going to do this one, because it seemed pretty simple, but thanks for taking it on.

jmle:
> This looks right to me. I was going to do this one, because it seemed pretty simple, but thanks for taking it on.

@gnu-andrew no problem! It would be great if you could take a look at [this PR](https://github.com/windup/windup-rulesets/pull/637) too, since I wasn't very sure about the hint wording.

rules-reviewed/openjdk11/openjdk8/tests/java-removals.windup.test.xml
```diff
@@ -54,6 +55,18 @@
...
+                    </iterable-filter>
+                </not>
+            </when>
+            <perform>
+                <fail message="[java-removals] `java-removals-00101-test failed"/>
+            </perform>
+        </rule>
+    </rules>
+</ruleset>
+        </ruletest>
```
carlosthe19916, line 72: These 2 last closing tags are not aligned. Could you please format them?

rules-reviewed/openjdk11/openjdk8/tests/data/java-removals/Base64Removals.java
```diff
@@ -0,0 +1,20 @@
...
+        BASE64Decoder base64Decoder = new BASE64Decoder();
+
+        try {
+            base64Decoder.decodeBuffer("inputString");
+            base64Encoder.encodeBuffer(new byte[]{'a', 'b'});
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+}
```
carlosthe19916, line 20: If I replace this file content with:
```
package com.test.project;

import sun.misc.*;

import java.io.IOException;

public class Base64Removals {
    public static void main(String[] args) {
        BASE64Encoder base64Encoder = new BASE64Encoder();

        try {
            base64Decoder.decodeBuffer("inputString");
            base64Encoder.encodeBuffer(new byte[]{'a', 'b'});
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

Then no Hint is being fired. I doubt this has to to with the Rule created in this PR but I'd like to understand/know where the issue is and how to fix it.

carlosthe19916 (commented):
@jmle The PR and Rule definition looks good to me. However, I've found an edge case where a Valid Java Class doesn't generate an appropriate Hint. 

Steps to reproduce the issue:
- Replace the content of the file `rules-reviewed/openjdk11/openjdk8/tests/data/java-removals/Base64Removals.java` by:

```
package com.test.project;

import sun.misc.*;

import java.io.IOException;

public class Base64Removals {
    public static void main(String[] args) {
        BASE64Encoder base64Encoder = new BASE64Encoder();

        try {
            base64Decoder.decodeBuffer("inputString");
            base64Encoder.encodeBuffer(new byte[]{'a', 'b'});
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

- Execute:
```
mvn clean test -DrunTestsMatching=java-removals
```
- Open the static report generated at `target/WindupRulesTests/windupgraph_UUID/reports/index.html` and go to the `Issues Tab`
- There is no Hint generated. I expect a hint to be generated since the Java class is a valid one and contains the line `BASE64Encoder base64Encoder = new BASE64Encoder()`

NOTE: the trick to generate the bug was to use an `import` that uses the `*` character. and also include only either BASE64Encoder or BASE64Decoder in the class.

rules-reviewed/openjdk11/openjdk8/java-removals.windup.xml
```diff
@@ -79,5 +79,23 @@
                 </hint>
             </perform>
         </rule>
+        <rule id="java-removals-00101">
+            <when>
+                <or>
+                    <javaclass references="sun.misc.BASE64Encoder"/>
+                    <javaclass references="sun.misc.BASE64Decoder"/>
+                </or>
```
mrizzi, line 87: I think it would be great to try to have just one condition with a `where` parameter to "inject" `Encoder` or `Decoder` and use the same also to drive the hint's message below.

mrizzi (commented):
Added a comment to suggest to leverage a `where` parameter.

carlosthe19916:
@jmle @mrizzi as discussed internally, I've created a Ticket here https://issues.redhat.com/browse/WINDUPRULE-832 so we can work on the bug in a separate PR.

I've tested this PR again and it works fine, even in the Edge case I reported previously in this PR. I've also found that lines like

```
BASE64Encoder base64Encoder = new BASE64Encoder();
```

Are going to generate twice the same hint. Eg.
Screenshot from 2022-03-25 10-43-52

And this "double hint" was, as discussed internally, acceptable. So I will approve this PR!

carlosthe19916 (approved):

PR #644, opened 2022-03-25
WINDUPRULE-800: add rule for unimplemented Thread.destroy() method

m-brophy:


PhilipCattanach:
Hi @m-brophy,
The change looks fine to me. The test works as well. I'm happy with what you have done, however there appears to be a conflict that we need to resolve. Thanks, Phil

jmle (approved):

PR #649, opened 2022-03-30
WINDUPRULE-820: rule for potential problems when constructing Bootstr…

m-brophy:
…ap Class Loader in JDK11


jmle:
Everything looks alright to me :+1:

PR #650, opened 2022-03-30
WINDUPRULE-788 - ComponentPeer removal rules

PhilipCattanach:
@jmle Can I ask you to investigate an issue for me?
The Componet.getPeer() aspect of this rule is not firing even though I have rebuild the CLI after the PR to add the Component.class to the removed classes jar.
If I run a custom rule from the 5.3.0 CLI  containing exactly the same logic. It does fire.


PhilipCattanach:
@jmle - Would you mind reviewing the PR now. I have updated the test. Thanks.

PR #651, opened 2022-03-31
Add DLL rule

agoncal:
Some Java projects can embed DLL files for native Windows calls.

This rule checks if there are *.dll files in the project


mrizzi (commented):
It looks good to me, I just have a doubt about the effort value I commented about.

PR #668, opened 2022-04-01
[WINDUPRULE-837] Update target rhr ruleset to reflect the springboot version range supported

jmle:
See [https://issues.redhat.com/browse/WINDUPRULE-837](https://issues.redhat.com/browse/WINDUPRULE-837).

### DIFF
--- a/rules-reviewed/rhr/springboot/springboot.rhamt.xml
+++ b/rules-reviewed/rhr/springboot/springboot.rhamt.xml
@@ -19,12 +19,12 @@
         <rule id="springboot-00001">
             <!-- rule condition, when it could be fired -->
             <when>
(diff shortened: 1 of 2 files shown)


PhilipCattanach:
@jmle Thanks- I'll give it a test

PhilipCattanach:
@jmle I've reviewed the changes and tested it successfully. Can I request a minor change to it please?
I think we only need 1 link for each of the rules. The link we need to retain is 
https://access.redhat.com/articles/3348731
I would remove the other 2 as they aren't adding any value. 
Thanks

PhilipCattanach:
Thanks @jmle That looks fine. I will merge it as soon as the automated integration tests pass.

PR #693, opened 2022-05-27
Spring boot to azure schedule job

agoncal:
Created a first Azure migration rule to determine whether application relies on scheduled jobs


mrizzi (approved):
@agoncal thanks a lot: really happy for this contribution :+1:

PR #694, opened 2022-05-30
Spring boot to azure database

agoncal:


rules-reviewed/technology-usage/embedded-framework-technology-usage.windup.xml
```diff
@@ -1036,5 +1036,19 @@
                 </technology-identified>
             </perform>
         </rule>
+        <rule id="technology-usage-embedded-framework-08900">
```
mrizzi, line 1039: This rule has been already committed in https://github.com/windup/windup-rulesets/pull/693

mrizzi (commented):
I think it's worth merging latest `master` because an ID seems wrong and also a rule has been already committed.

PR #700, opened 2022-06-02
EAP to Azure AppService from Maven projects

dfuenzalida:


rules-reviewed/azure/eap/eap-to-azure-appservice-pom.windup.xml
```diff
@@ -0,0 +1,68 @@
...
+             <addon id="org.jboss.windup.rules,windup-rules-xml,3.0.0.Final"/>
+        </dependencies>
+        <sourceTechnology id="eap" versionRange="[7,8)" />
+        <targetTechnology id="azure-appservice"/>
+        <tag>Azure</tag>
+    </metadata>
+    <rules>
+        <rule id="eap-to-azure-appservice-pom-001">
+            <when>
+                <file filename="pom.xml" />
```
mrizzi, line 21: I think it's worth investing some effort to have a more restrictive condition to identify an "EAP application" otherwise the rule will become very noisy.
An idea could be identify the usage of the `org.jboss.bom` artifact and more info available https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.4/html-single/development_guide/index#manage_project_dependencies

mrizzi (commented):
Suggested an improvement to the rule's condition to make it more effective in identifying an "EAP application"

rules-reviewed/azure/eap/eap-to-azure-appservice-pom.windup.xml
dfuenzalida: @mrizzi I've updated the PR to use the JBoss BOM with a specific version that would match the minimum version of JBoss offered in App Service

m-brophy (approved):

PR #701, opened 2022-06-03
Fix the data path

agoncal:


mrizzi (approved):
Thanks for improving the test scenarios :+1:

PR #702, opened 2022-06-03
Azure rule for file system

agoncal:


mrizzi (approved):
Thanks a lot @agoncal :+1:

PR #703, opened 2022-06-03
Fix the Azure URL documentation

agoncal:


mrizzi (approved):
Works fine, thanks @agoncal :+1:

PR #704, opened 2022-06-03
Azure rule for read-only static content

agoncal:


rules-reviewed/azure/springboot/spring-boot-to-azure-static-content.windup.xml
```diff
@@ -0,0 +1,45 @@
...
+    <rules>
+        <rule id="spring-boot-to-azure-static-content-01000">
+            <when>
+                <or>
+                    <file filename="{*}.htm"/>
+                    <file filename="{*}.html"/>
+                    <file filename="{*}.css"/>
+                    <file filename="{*}.scss"/>
+                    <file filename="{*}.js"/>
+                </or>
```
mrizzi, line 27:
```suggestion
                <file filename="{*}.{extensions}"/>
```

mrizzi (commented):
The rule, as it is, works fine but I proposed an enhancement to just have one condition evaluated against a `where` parameter.

mrizzi (approved):
Nice contribution, thanks @agoncal :+1:

PR #705, opened 2022-06-03
Azure rule for clear password

agoncal:

PR #709, opened 2022-06-11
Look for plain-text passwords in config files

dfuenzalida:
Based on #705 with an extra entry for XML files


mrizzi (commented):
I have some ideas about trying to improve this rule and also the one you mentioned coming from #705 but I need some time to do the changes a push a PR.
I'll ping you as soon as it'll be available.

PR #712, opened 2022-06-17
EAP to Azure AppService - Check usage of filesystem APIs

dfuenzalida:


PhilipCattanach:
Hi @dfuenzalida 
This ruleset is almost identical to what @agoncal implemented for the springboot source in PR https://github.com/windup/windup-rulesets/pull/702/files
To save duplication I have moved rules that are applicable to both sources and targets to the rules-reviewed/azure folder rather than at the rules-reviewed/azure/SOURCE level in PR https://github.com/windup/windup-rulesets/pull/714
It is a more concise way of implementing rules that are applicable to both sources and targets.
Happy to discuss if you have any questions.

PR #714, opened 2022-06-17
WINDUPRULE-863 - Azure file system for 2 sources and 2 targets

PhilipCattanach:
There were separate rulesets submitted, that were almost identical, for checking for file system io.
The only real difference was the source.
To save duplication I have moved rules that are applicable to both sources and targets to the rules-reviewed/azure folder rather than at the rules-reviewed/azure/SOURCE level
It is a more concise way of implementing rules that are applicable to both sources and targets.


mrizzi (approved):
The refactor makes a lot of sense, thanks @PhilipCattanach :+1:

PR #721, opened 2022-06-23
EAP to Azure AppService - Look for datasource drivers in config files

dfuenzalida:
Look for database drivers in configuration files, since these need a script to be installed during application startup


rules-reviewed/azure/eap/eap-to-azure-appservice-datasource-driver.windup.xml
```diff
@@ -0,0 +1,44 @@
...
+        <sourceTechnology id="eap" versionRange="[7,8)" />
+        <targetTechnology id="azure-appservice"/>
+        <targetTechnology id="azure-aks"/>
+        <tag>password</tag>
+        <tag>security</tag>
+    </metadata>
+    <rules>
+        <rule id="eap-to-azure-appservice-datasource-driver-01000">
+            <when>
+                <xmlfile matches="//driver" />
```
m-brophy, line 22: I think this is too broad a filter. Searching for a `<driver>` element will find non-database related xml elements too, for instance libvirt disk config xml has a driver element. Maybe we can add a pattern for the acceptable contents of the element ( eg postgresql|jdbc|h2|mongodb etc)?

m-brophy (changes requested):

m-brophy (approved):

PR #723, opened 2022-06-30
Winduprule 878

PhilipCattanach:
The way to test this change is building the cli with this Rules PR applied to it
Then analyze the the spring petclinic jar [spring-petclinic-2.4.0.BUILD-SNAPSHOT.jar]
Look through list of potential issues and there should be numerous issues with the title
"Spring component whatever requires investigation"
If you click on that issue it is generated from an embedded spring*.jar file and if you click on the Show Rule it will show springboot-generic-catchall-00100.

Then go down to the Information section of the report.
There will be numerous Information issues pointing at Spring artifacts.
Prior to this PR what was happening was artifacts that were causing the Information category issue rules to fire were creating a hint against the artifact, that would prevent the springboot-generic-catchall-00100 rule from firing (it will only fire if there is not already a hint associated to that artifact).

Now the springboot-generic-catchall ruleset runs in an earlier phase so there will be a Potential issue for every embedded spring jar and a Information issue for most of them. There should never be a Information issue for a bonafide spring artifact without a corresponding issue for the same artifact in the Mandatory or Potential sections of the report.

Equivalent changes are not required to the springboot-generic-catchall-groovy-00000 rule as this attaches the hint to dependency declared within the pom.xml not the embedded spring jar file.

### DIFF
--- a/rules-reviewed/technology-usage/3rd-party.windup.xml
+++ b/rules-reviewed/technology-usage/3rd-party.windup.xml
@@ -10,6 +10,8 @@
             <addon id="org.jboss.windup.rules,windup-rules-javaee,3.0.0.Final" />
             <addon id="org.jboss.windup.rules,windup-rules-java,3.0.0.Final" />
         </dependencies>
+        <phase>PostMigrationRulesPhase</phase>
+        <executeBefore>technology-usage-3rd-party</executeBefore>
     </metadata>
     <rules>
         <rule id="3rd-party-01000">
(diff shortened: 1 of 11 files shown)


mrizzi (commented):
After some tests using CLI latest snapshot (2022-07-13), it has been verified there's no actual interaction between the rules because of the `springboot-generic-catchall` rule being based on hint existence and the "tech" rules adding classifications (and not hints).
So we decided to close this PR without merging.

PR #760, opened 2022-10-10
WINDUPRULE-892: new jakarta rule for empty beans.xml file

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-892

with thanks to @bstansberry, based on his rule at https://github.com/bstansberry/windup-rulesets/commit/769eb0f1a3d8ab2674cfb998945e42202e220a2a


PhilipCattanach:
Hi Mark, I tested the rule with the test file and a non-empty beans.xml file and it fired with the empty file and it didn't fire with the populated file. 
The only thing that I think needs to change is to add a sourceTechnology tag. This is applicable for users upgrading from eap6 onwards.

PR #770, opened 2022-10-20
WINDUPRULE-893: eap8 ejb removed methods rules

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-893


rules/rules-reviewed/eap8/eap6/eap8.ejb.windup.xml
```diff
@@ -0,0 +1,82 @@
...
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,3.0.0.Final" />
+            <addon id="org.jboss.windup.rules,windup-rules-java,3.0.0.Final" />
+        </dependencies>
+        <sourceTechnology id="eap" versionRange="[6,7)" />
+        <targetTechnology id="eap" versionRange="[8,)" />
+        <tag>eap8</tag>
+    </metadata>
+    <rules>
+        <rule id="eap8-ejb-00001">
+            <when>
```
jmle, line 19: According to the comment on the ticket:
> So the rules need to look for method calls below, on a class that implements the above interface.

I think the rules are missing the `on a class that implements the above interface` part. So I think it should be something like this:

```
                <javaclass references="EJBContext">
                    <location>IMPLEMENTS</location>
                </javaclass>
                <javaclass references="javax.ejb.EJBContext.getCallerIdentity({*})" from="default">
                    <location>METHOD_CALL</location>
                </javaclass>
```

rules/rules-reviewed/eap8/eap6/eap8.ejb.windup.xml
```diff
@@ -0,0 +1,82 @@
...
+                <hint title="javax.ejb.EJBContext.getEnvironment() method removed" effort="1" category-id="mandatory">
+                    <message>Method `javax.ejb.EJBContext.getEnvironment()` was deprecated and has now been removed from the Jakarta Enterprise Beans 4.0 specification and api.
+                        Use the JNDI naming context `java:comp/env` to access enterprise bean's environment.</message>
+                    <link title="Jakarta Enterprise Beans 4.0 specification"
+                          href="https://jakarta.ee/specifications/enterprise-beans/4.0/"/>
+                    <tag>eap8</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="eap8-ejb-00004">
```
jmle, line 66: Would it be possible to squeeze all these rules into a single rule with `<or>`?

jmle (changes requested):

rules/rules-reviewed/eap8/eap6/eap8.ejb.windup.xml
m-brophy: The ticket also mentions this
> WINDUP's ([Javadoc](http://windup.github.io/windup/docs/latest/javadoc/org/jboss/windup/ast/java/data/TypeReferenceLocation.html)) METHOD_CALL location qualifier works transitively with interfaces. So the rules can be coded to look for
>FullyQualifiedInterfaceName.MethodName()
>And any concrete classes that have implemented the interface and calling the method will be identified.

So I don't think we need to make those changes. wdyt?

jmle: Isn't it possible that the METHOD_CALL is being made from outside the class, even with a fully qualified name?

m-brophy: I thought the point made in the ticket was that the METHOD_CALL qualifier will pick up calls on classes implementing the interface too -  there's no other way of calling these methods is there?

PR #772, opened 2022-11-02
Jakarta EE 10 dependencies openrewrite recipe

m-brophy:


m-brophy:
Hi @PhilipCattanach -

The version value is an actual value - 2.1.X just means the latest minor version of the 2.1 version set, openrewrite works it out itself.

The ChangeDependencyScope call removes the scope from the dependency if you don't provide it with a scope parameter so that's what that is doing. The reason it's there is that to migrate an app you first run the existing JavaxToJakarta recipe so that the imports etc get updated correctly but that provides a new dependency to fit the altered code,  `jakarta.annotation-api` in the example you give. When we then run this recipe it converts the jboss dependency to the same artifact. However if it's a starter-app the scope is `provided`. We make a call to remove duplicate dependencies like this at the end of the recipe, but this doesn't do anything if the scope is  different on the two dependencies so we delete the scope after updating jboss javax dependencies before deleting duplicates.

PR #774, opened 2022-11-07
[WINDUPRULE-913] Temporary fix for PKIX cert issue

jmle:
See [JIRA ticket](https://issues.redhat.com/browse/WINDUPRULE-913).


carlosthe19916:
Thanks @jmle tested locally and it the temporary workaround works. We should continue working on a more permanent solution but for now I will merge this PR so we have our pipelines working again.

PR #785, opened 2022-11-29
[WINDUPRULE-896] Jakarta servlet related rules

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-896).


PhilipCattanach:
HI @jmle  - Nearly there, Just a couple of minor changes please.
Rule 00071 and 00072 can you add the signature of the methods removed
Rule 00102 has the method name missing from the title
Rule 00112 please provide the signature of the new method in the message
Please change any rules with 2 SP to 3 SP
See section A3.2 in this guide about the SP values
https://access.redhat.com/documentation/en-us/migration_toolkit_for_applications/6.0/html/cli_guide/reference_material

PR #794, opened 2022-12-06
Detects Java SE Threads usage

agoncal:
This rule detects that an application is using Java SE Thread APIs. This is important to detect when running in an application server or the cloud


PhilipCattanach:
Many thanks @agoncal - I'll review your PR.

PR #824, opened 2022-12-13
[WINDUPRULE-903] Keycloak/OpenID rules

jmle:
[Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-903)


PhilipCattanach:
Hi @jmle 
There appears to be  2 rules required and there is only one rule in the ruleset. The rule and test for the presence of a keycloak.json file is missing, even though there is a keycloak.json in the tests/data folder.
The migration document states that this rule is applicable to source EAP 6 and above, but the sourceTechnology tag doesn't reflect this.
With Windup the `string` text highlighting doesn't work in the title so perhaps remove it. But I would highlight `auth-method` and `OIDC` in the message.
Thanks,
Phil

PhilipCattanach:
Tested successfully

PR #840, opened 2023-01-13
WINDUPRULE-942 allow eap7 source

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-942


mrizzi (approved):
Tested locally with `--source eap7 --target eap8` and the rules worked.
Thanks @m-brophy :+1:

PR #846, opened 2023-02-01
Tag versions of Java EE based on deployment descriptors

agoncal:
These rules check for the version number of Java EE deployment descriptors and tags the technology used (eg. EJB) and its version (eg. 3.2)


agoncal:
Wait for https://github.com/windup/windup/pull/1587

m-brophy:
Running an analysis with the latest version of the rules against the test objects produces this index page:
image
 (EJB, JPA  and Servlet xml rules are now not needed because the java rules which should have previously provided the output required have been fixed)

m-brophy:
running against the previous version of test data files (ie including EJB, JPA and Servlet test data files) produces this output (note the tags produced by the java rules):
image

agoncal:
@m-brophy @mrizzi really good !!! 

So, is it ready to be merged ?

mrizzi (commented):
@m-brophy I've run a test and in technolgies report I can see all the new tech tags but `JSF XML`: can you add it as well please?

PR #848, opened 2023-02-06
Fixing URLs

agoncal:
These 3 URLs return a 308 because they do not have a `/` at the end. I'm adding the missing `/` so it returns a 200 instead


mrizzi (approved):
LGTM, thanks @agoncal :+1:

PR #859, opened 2023-02-16
[WINDUPRULE-902] Add picketlink ruleset

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-902).


rules/rules-reviewed/eap8/eap7/picketlink.windup.xml
```diff
@@ -0,0 +1,34 @@
...
+            <addon id="org.jboss.windup.rules,windup-rules-javaee,3.0.0.Final"/>
+            <addon id="org.jboss.windup.rules,windup-rules-java,3.0.0.Final"/>
+        </dependencies>
+        <targetTechnology id="eap" versionRange="[8,9)"/>
+    </metadata>
+    <rules>
+        <rule id="picketlink-00010">
+            <when>
+                <or>
+                    <file filename="{*}WEB-INF/{*}picketlink.xml"/>
```
carlosthe19916, line 19: This particular line does not work. I've verified it this way:
- Keep only the condition `<file filename="{*}WEB-INF/{*}picketlink.xml"/>` and remove the other conditions so the final value is:

```
                <or>
                     <file filename="{*}WEB-INF/{*}picketlink.xml"/>
                </or>
```

- Execute the test `mvn test -DrunTestsMatching=picketlink.windup.test.xml`
- The test should fail, which is expected, because the number of hints expected don't match. Just ignore it.
- Open the generated report at `rules/target/WindupRulesTests/windupgraph_RANDON_STRING/reports`
- In your browser navigate to the page `Issues` and you will see that the file `picketlink.xml` did not generated any issue despite the fact that that file is located at the folder `WEB-INF` as the rule expects

rules/rules-reviewed/eap8/eap7/picketlink.windup.xml
```diff
@@ -0,0 +1,34 @@
...
+                    <filecontent pattern="picketlink" filename="{*}.xml"/>
+                    <file filename="picketlink-sts.xml"/>
+                    <filecontent pattern="PicketLinkSTS" filename="picketlink.xml"/>
```
carlosthe19916, line 22: I have noticed the current rule does not contain links for the user to understand better the issue.
From the document attached to the ticket I get this info:

_If a picketlink-sts.xml file or occurrences of “PicketLinkSTS” were detected in the picketlink.xml file, the user will need to configure Apache CXF STS instead as described briefly [here](https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.4/html-single/developing_web_services_applications/index#ws_trust_sts_apache_CXF_support). (If more details are needed here, we’ll need to ask the Web Services team to provide this information.)_

Is there a reason for not including the link and described above?

carlosthe19916 (changes requested):
@jmle Thanks for the PR. I've left a couple of comments, could you please take a look at them?

carlosthe19916 (approved):

PR #866, opened 2023-02-21
WINDUPRULE-897: Rules to move to Jakarta XML Binding 4.0

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-897

### DIFF
--- a/rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
+++ b/rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
@@ -153,13 +153,12 @@
         <rule id="eap8-xml-binding-00007">
             <when>
                 <or>
-                    <javaclass references="javax.xml.bind.Marshaller.setAdapter({*})">
-                    </javaclass>
+                    <javaclass references="javax.xml.bind.Marshaller.setAdapter({*})"/>
                 </or>
             </when>
             <perform>
-                <hint title="Implementation or use by the application of the `javax.xml.bind.Marshaller` `setAdapter()` methods may need to be updated" effort="1" category-id="mandatory">
-                    <message>Implementation or use by the application of the `javax.xml.bind.Marshaller` `setAdapter()` methods may need to be updated
+                <hint title="Implementation or use by the application of the javax.xml.bind.Marshaller setAdapter() method may need to be updated" effort="1" category-id="potential">
+                    <message>Implementation or use by the application of the `javax.xml.bind.Marshaller` `setAdapter()` method may need to be updated
                         to conform to the more fully specified generic type signatures in Jakarta Xml Binding 4.0.</message>
                     <link title="Jakarta XML Binding 4.0 specification"
                           href="https://jakarta.ee/specifications/xml-binding/4.0/jakarta-xml-binding-spec-4.0.html"/>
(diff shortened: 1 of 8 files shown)


rules/rules-reviewed/eap8/eap6/tests/data/xml-binding/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.html</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.glassfish.jaxb</groupId>
+            <artifactId>jaxb-core</artifactId>
```
jmle, line 23: I can't see a rule for the detection of this dependency

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
```diff
@@ -0,0 +1,161 @@
...
+        <sourceTechnology id="eap" versionRange="[6,7]" />
+        <targetTechnology id="eap" versionRange="[8,)" />
+        <tag>eap8</tag>
+    </metadata>
+    <rules>
+        <rule id="eap8-xml-binding-00001">
+            <when>
+                <or>
+                    <javaclass references="com.sun.xml.bind.api.{*}">
+                        <location>IMPORT</location>
```
jmle, line 22: What if the class is used with the full qualified name instead of importing? Maybe it's better to remove the `IMPORT` location in this one and similar ones?

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
```diff
@@ -0,0 +1,161 @@
...
+                        Update package paths  from `com.sun.xml.bind.*` to `org.glassfish.jaxb.runtime.*`</message>
+                    <link title="Jakarta XML Binding 4.0 specification"
+                          href="https://jakarta.ee/specifications/xml-binding/4.0/"/>
+                    <tag>eap8</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="eap8-xml-binding-00002">
+            <when>
+                <javaclass references="javax.xml.bind.Marshaller.setProperty({*})" as="setPropertiesMethodCalls">
```
jmle, line 41: Doesn't this work with just `<javaclass references="javax.xml.bind.Marshaller.setProperty({property})">...` and then a `<where param="property"><matches pattern="com.sun.xml.bind{*}"></where>`?

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
```diff
@@ -0,0 +1,161 @@
...
+                          href="https://jakarta.ee/specifications/xml-binding/4.0/"/>
+                    <link title="Implementation discovery in Jakarta XML Binding 4.0"
+                          href="https://jakarta.ee/specifications/xml-binding/4.0/jakarta-xml-binding-spec-4.0.html#implementation-discovery"/>
+                    <tag>eap8</tag>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="eap8-xml-binding-00005">
+            <when>
+                <javaclass references="javax.xml.bind.JAXBContext.newInstance({*})" as="instanceMethods">
```
jmle, line 97: Will this rule detect particular keys in the map passed?

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
```diff
@@ -0,0 +1,161 @@
...
+                        Passing a schema to `jakarta.xml.bind.Marshaller.setSchema(Schema)` or  `jakarta.xml.bind.UnMarshaller.setSchema(Schema)` is the replacement.</message>
+                    <link title="Jakarta XML Binding 4.0 specification"
+                          href="https://jakarta.ee/specifications/xml-binding/4.0/"/>
+                </hint>
+            </perform>
+        </rule>
+        <rule id="eap8-xml-binding-00007">
+            <when>
+                <or>
+                    <javaclass references="javax.xml.bind.Marshaller.setAdapter({*})">
```
jmle, line 147: I think we're missing `getAdapter()` somewhere around here

jmle (changes requested):

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
m-brophy: I don't think it would because the method parameters are specified in the rule as a type reference not as  a particular value so you'd have to specify (java.lang,String, java.lang.Object) which doesn't drill down to the value we want.

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
m-brophy: This one is all to do with a change in the generics accepted as parameters in these methods. Looking at the Marshaller api docs,  getAdapter accepts a parameter  (Class<A> type)  so you may well be right. However it wasn't mentioned in the spec document!

m-brophy: I've asked the question on the JIRA

rules/rules-reviewed/eap8/eap6/eap8-xml-binding.windup.xml
m-brophy: Not exactly no, it detects classes using these instantiation methods and searches within those classes for the offending class names. So it will detect the desired conditions but I imagine there will be some false positives on occasion.

jmle (approved):

PR #881, opened 2023-02-27
Restrict IO classes to only Files

agoncal:
The rules has too many fault postives (eg. IOExcpetion or Serializable). This new rule only focuses on File access


mrizzi (approved):
LGTM, thanks for the enhancement @agoncal :+1:

PR #889, opened 2023-03-09
WINDUPRULE-955: updating spring boot tech rhr supported versions

m-brophy:
https://issues.redhat.com/browse/WINDUPRULE-955


PhilipCattanach:
@m-brophy - You have included the spring-boot-jarmode-layertools-2.7.9.jar in the test data.
It is a genuine SB artifact with the same SHA1 hash as artifact in Maven Central.
Our CPaaS build systems will recognize it and flag it as a dependency to windup. Then CVEs associated to it will be linked to WINDUP. Have a word with Carlos about how to replace this true file with a dummy one so it isn't recognized as the true SB artifact. Thanks.

PR #890, opened 2023-03-09
add version and correct a version for deps in pom.xml so it runs

shawn-hurley:


shawn-hurley:
@jmle if you can take a look, this will help us with the tests that we are running for parity testing.

shawn-hurley:
Hey folks, 

I apologize for the PR with no context; that is not very cool. 

This change aims to make the `pom.xml` file parsable by maven. 

We currently use the pom.xml and maven to determine the dependencies by running mvn command; as our work on using the testing directories to validate feature parity, it would be helpful for the pom file to be valid. 

I hope that context helps. Let me know if this is not a welcome PR or if there is some other option :) 

Thanks,
Shawn Hurley

mrizzi:
Hi @shawn-hurley,
thanks for the context.

TBH, I don't think, within the "feature parity" landscape, it's recommended to change test data.
I think this file spotted a difference between the tools:

- Windup manages `pom.xml` files with the only requirement of them being formally compliant with schema
- new analyzer has a further requirement for `pom.xml` files to be "buildable" by Maven

As a kind of side note, I think it also highlights another difference:

- Windup works "offline" without requiring any on line data
- new analyzer needs go be on line for running Maven command

So, even if I fully understand the good in these specific changes you're suggesting, I would like to better understand what's the approach we're following in this feature parity task.
For example I think it would valuable to collect the differences between the tools to understand if and how they're acceptable rather than running the risk of changing the test data and forget about the differences.

Please, let me know you're view on the overall approach to changing test data vs feature parity goal.

mrizzi (approved):
I think we can follow up the discussion from my comment above out of this PR because, as I wrote, I can see the good in these changes to be merged.

Thanks @shawn-hurley and congrats on your first contribution :tada:

PR #903, opened 2023-03-17
[WINDUPRULE-959] Add missing links

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-959).


PhilipCattanach:
@jmle Is the link title and URL in the legacy-vault ruleset correct?
I'm not seeing any information (unless I'm missing something) that helps the use understand more about this change?

PR #557, opened 2021-05-10
Added support for dotnet cloud-readiness analysis

vinodpahuja:

PR #904, opened 2023-03-17
Added support for dotnet cloud-readiness analysis

vinodpahuja:
ruleset-explorer

Continuing PR #557 here.
- Rebased fork
- Added test

### DIFF
--- a/rules/rules-reviewed/openshift/dotnet/db2-unmanaged.windup.xml
+++ b/rules/rules-reviewed/openshift/dotnet/db2-unmanaged.windup.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0"?>
+<ruleset id="db2-unmanaged"
+    xmlns="http://windup.jboss.org/schema/jboss-ruleset"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://windup.jboss.org/schema/jboss-ruleset http://windup.jboss.org/schema/jboss-ruleset/windup-jboss-ruleset.xsd">
+    <metadata>
+        <description>
+            This is a ruleset for .net db2-unmanaged topics when migrating an application to cloud environments.
+        </description>
+        <dependencies>
+            <addon id="org.jboss.windup.rules,windup-rules-base" />
+        </dependencies>
+        <sourceTechnology id="dotnet"/>
+        <targetTechnology id="cloud-readiness"/>
+        <tag>db2-unmanaged</tag>
+    </metadata>
+    <rules>
+        <rule id="db2-unmanaged-0000">
+            <when>
+                <filecontent filename="{*}.{cs|vb}" pattern="IBM.Data.DB2"/>
+            </when>
+            <perform>
+				<hint title="Unmanaged DB2 Database" effort="7" category-id="cloud-mandatory">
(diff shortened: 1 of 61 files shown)


vinodpahuja:
@PhilipCattanach @mrizzi - Please review and suggest further.

PhilipCattanach:
Hi Vinod,
Thank you for you latest changes. Here is some more detailed feedback, that requires further changes to the PR:

We want to have these rules laser focused on the use case that they support to that end we need to change the target of the rules to dotnet-cloud 

The folder structure needs to change to rules-reviewed/dotnet-cloud/dotnet

All of the windup.xml and windup.test.xml file names should all be in lower case

Each ruleset needs to be atomic, so each ruleset has a dedicated data subfolder containing the test data that drives that ruleset.
For example the windowsAuth.windup.xml (rename to say windows-auth.windup.xml) should have a data folder
rules-reviewed/dotnet-cloud/dotnet/data/windows-auth
And the path to the data folder must be reflected in the testDataPath tag in the corresponding windup.test.xml file

The message portion of the rule is often just the title repeated.
Look at the request-filter rule.
The messages need to elaborate on what the issue is and how it can be addressed.

The msmq.windup.xml has a message “Using System.Messaging”

There are very few links. Within the rules Links are extremely useful for providing supporting information/points of reference.
 
Rules within these that highlight potential issues e.g. transactions.windup.xml, in this use case (moving to cloud) should be cloud-optional (there isn’t a cloud-potential category).

The use of tags is very inconsistent
In some case you have database-access in others DatabaseAccess

I would suggest you review the existing set of tags already available in windup.
The way to do that is to download the CLI from https://windup.github.io/downloads/
Unzip it
Navigate to the bin folder and in the command window run the windup-cli script with the --listTags argument.

vinodpahuja:
Thanks @PhilipCattanach for detail feedback, will work on it.

rules/rules-reviewed/openshift/dotnet/db2-unmanaged.windup.xml
```diff
@@ -0,0 +1,31 @@
...
+            <addon id="org.jboss.windup.rules,windup-rules-base" />
+        </dependencies>
+        <sourceTechnology id="dotnet"/>
+        <targetTechnology id="cloud-readiness"/>
+        <tag>db2-unmanaged</tag>
+    </metadata>
+    <rules>
+        <rule id="db2-unmanaged-0000">
+            <when>
+                <filecontent filename="{*}.{cs|vb}" pattern="IBM.Data.DB2"/>
```
shawn-hurley, line 20: We are considering adding multi-language support using https://github.com/konveyor/analyzer-lsp and adding another provider. One of the things that we use is the language server protocol to allow for searching source code based on that language's server. 

I would be interested in if there were some subset of rules that we could work together if you are interested.

shawn-hurley: You can find us on konveyor-dev slack channel in k8s slack or you can email me :) looking forward to hearing from you!

vinodpahuja: @shawn-hurley - Lets connect and see how we can compliment each other.
Please share your contact, I am available at vinod.pahuja@gmail.com

PhilipCattanach:
I am going to close this PR as it is more appropriate for Konveyor than Windup

PR #907, opened 2023-03-21
[WINDUPRULE-971] Fix and improve hibernate6-00270 rule

jmle:
- Simplified rule in order to improve performance
- Made rule `potential`
- Added mention to the fact that the new artifact that contains the dialects is not included in EAP

See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-971).


PhilipCattanach:
The changes look fine to me and I have ran a successful test.
Thanks,
Phil

PR #923, opened 2023-04-17
Add new runtime label for Spring Boot Apps

showpune:
Add New runtime lable for spring boot:
It is copied from runtime label tomcat, 
1) Add the component like JMS, JavaMail, Java-WS, JPA as SpringBoot can integrate these component by starter
2) EJB Component, JCA and JTA component is still forbidden
3) Add some GUI interface like eclipse RCP, JNLP and Applet as unsuitable


showpune:
@agoncal , add the runtime lable for spring boot

PR #937, opened 2023-05-01
Changing name of same exact class to have different names

shawn-hurley:
This is another small fix, that allows for the determining of correct information to be loaded during the web sphere tests.


PhilipCattanach:
@shawn-hurley Thanks for the PR.
Just for info anything under rules/rules-archived are not included in the shipped product.
It is a just a folder that allows us to keep for reference rules that are no longer required.
Typically for obsolete migration targets such as EAP 6

PR #953, opened 2023-05-15
add azure redis connection string config rule

KaiqianYang:


PhilipCattanach:
Successfully reviewed and tested.

PR #957, opened 2023-05-17
check java version for azure

KaiqianYang:


KaiqianYang:
@PhilipCattanach Hi Philip, in this commit I encountered a issue of [xmlfile match not working](https://issues.redhat.com/browse/WINDUP-3872). I also tested many other xmlfile matching patterns but all didn't work. 

Could you help me with this issue? Thank you!

PhilipCattanach:
Hi @KaiqianYang 
I will take a look.

PhilipCattanach:
Created https://issues.redhat.com/browse/WINDUPRULE-995 to manage this change

KaiqianYang:
> Created https://issues.redhat.com/browse/WINDUPRULE-995 to manage this change

Hi Philip, thanks for your help!

I've merged your commit into this pr. Based on your version, I made an update to simplify the rule pattern by utilizing windup:matches().

I noticed that a key change you introduced was using /m:java.version instead of /java.version, and adding <namespace prefix="m" uri="http://maven.apache.org/POM/4.0.0"/> within <xmlfile>. Could you please provide more insight into this alteration? It seems different from standard XPath patterns. Additionally, is including <namespace prefix="m" uri="http://maven.apache.org/POM/4.0.0"/> a requirement for the rule to function properly?

Thank you

PhilipCattanach:
@KaiqianYang - thanks for the feedback. I'll take another look. I may have made a mistake.

PhilipCattanach:
@KaiqianYang - I'll ask one of my Engineers to get involved as their technical skills are so much better than my own.

Hi @jmle Can I ask you to review this  PR please. I have tried to help but my technical skills have been found wanting.

KaiqianYang:
> Hey @KaiqianYang, the `<xmlfile>` condition will need the `<namespace>` element if the given XML is expected to have namespace information. If not, the `<namespace>` element must be removed.
> 
> Therefore, if you don't know whether the XML to analyse will have namespace information or not, the best might be to do something like this:
> 
> ```
> <or>
>   <xmlfile matches="X"...>
>     <namespace ...>
>   </xmlfile>
>   <xmlfile matches="X" .../>
> </or>
> ```
> 
> This way, you will make sure that the condition is triggered regardless of whether the XML is "namespaced" or not.

Hi @jmle , thanks for your clarification

> Hey @KaiqianYang, the `<xmlfile>` condition will need the `<namespace>` element if the given XML is expected to have namespace information. If not, the `<namespace>` element must be removed.
> 
> Therefore, if you don't know whether the XML to analyse will have namespace information or not, the best might be to do something like this:
> 
> ```
> <or>
>   <xmlfile matches="X"...>
>     <namespace ...>
>   </xmlfile>
>   <xmlfile matches="X" .../>
> </or>
> ```
> 
> This way, you will make sure that the condition is triggered regardless of whether the XML is "namespaced" or not.

@jmle Thanks for your information. And I wanted to confirm that in Windup rule XML path expressions, the working pattern is `//m:java.version` instead of `//java.version`. This is different from standard XPath expressions. So, does this mean that specifying the namespace is necessary when using XML path patterns in Windup rules?

KaiqianYang:
> Hey @KaiqianYang, the `<xmlfile>` condition will need the `<namespace>` element if the given XML is expected to have namespace information. If not, the `<namespace>` element must be removed.
> 
> Therefore, if you don't know whether the XML to analyse will have namespace information or not, the best might be to do something like this:
> 
> ```
> <or>
>   <xmlfile matches="X"...>
>     <namespace ...>
>   </xmlfile>
>   <xmlfile matches="X" .../>
> </or>
> ```
> 
> This way, you will make sure that the condition is triggered regardless of whether the XML is "namespaced" or not.

@jmle Thanks for your information. And I wanted to confirm that in Windup rule XML path expressions, the working pattern is `//m:java.version` instead of `//java.version`. This is different from standard XPath expressions. So, does this mean that specifying the namespace is necessary when using XML path patterns in Windup rules?

jmle:
@KaiqianYang the namespace in the condition is necessary when the "target" XML has namespace information. If no namespace information is present in the XML, the namespace must be removed from the condition:

- Target XML with namespace :arrow_right: condition must have namespace
- Target XML without namespace :arrow_right: condition must not have namespace

Therefore, if you don't know if the XMLs you're targetting will or will not have the namespace information on them, the best you can do is to `<or>` two conditions as explained above.

KaiqianYang:
> @KaiqianYang the namespace in the condition is necessary when the "target" XML has namespace information. If no namespace information is present in the XML, the namespace must be removed from the condition:
> 
> * Target XML with namespace ➡️ condition must have namespace
> * Target XML without namespace ➡️ condition must not have namespace
> 
> Therefore, if you don't know if the XMLs you're targetting will or will not have the namespace information on them, the best you can do is to `<or>` two conditions as explained above.

Yes, your response is quite clear. Thank you.

rules/rules-reviewed/azure/azure-java-version.windup.xml
```diff
@@ -0,0 +1,67 @@
...
+        <targetTechnology id="azure-appservice"/>
+        <targetTechnology id="azure-aks"/>
+        <targetTechnology id="azure-container-apps"/>
+        <tag>version</tag>
+    </metadata>
+    <rules>
+        <rule id="azure-java-version-01000">
+            <when>
+               <xmlfile matches="//m:java.version[windup:matches(text(), '{v}')]" in="pom.xml" as="result">
+                        <namespace prefix="m" uri="http://maven.apache.org/POM/4.0.0"/>
```
jmle, line 25: Continuin with my comments, in this case, this is correct, since the POM files in the test folders have namespace information, although [it is possible to find POM files without namespace information that build correctly](https://stackoverflow.com/questions/25996084/pom-xml-namespace-necessary). If you want to capture those edge cases, you will have to `<or>` this condition with an additional one _without_ the `namespace` element.

jmle:
@KaiqianYang I was taking a look at this again. After my clarification, is there something you want to change or are you waiting for it to get merged?

KaiqianYang:
> @KaiqianYang I was taking a look at this again. After my clarification, is there something you want to change or are you waiting for it to get merged?

Hi @jmle , I think it's fine now. Waiting for it to get merged.
Thank you.

PR #970, opened 2023-05-25
detect aws configurations

KaiqianYang:


agoncal:
@mrizzi @PhilipCattanach couldn't that be a Technology Usage rule that could be targeted at discovery?

We could even have a few Technology Usage rules that could discover Azure SDK APIs. WDYT ?

PhilipCattanach:
@agoncal - Yes, the technologies discovered by these rules are very much the technologies that we want to 'discover' (surface as technology tags) and categorize so they appear in the appropriate tiers of the Technology Reports

agoncal:
We could have a set of rules for AWS, GCP, Azure, etc. inside `rules/rules-reviewed/technology-usage/cloud-providers.windup.xml`. WDYT ?

KaiqianYang:
> We could have a set of rules for AWS, GCP, Azure, etc. inside `rules/rules-reviewed/technology-usage/cloud-providers.windup.xml`. WDYT ?

@agoncal Yes, it’s a good idea. 

One additional thing is that, in technology-usage, it mainly **detects jar files**, e.g., <file filename="{*}log4j{*}.jar"/>, But the rule I made for AWS **detects configurations**. This is useful when customers are scanning source code instead of jars. We also have rules in Azure folder that detect technology usage by dependencies in Maven, configurations, and API usage. Should we also add or move them into technology-usage? e.g., [swing](https://github.com/windup/windup-rulesets/blob/master/rules/rules-reviewed/azure/springboot/spring-boot-to-azure-swing.windup.xml), [database](https://github.com/windup/windup-rulesets/blob/master/rules/rules-reviewed/azure/springboot/spring-boot-to-azure-database.windup.xml), [eureka](https://github.com/windup/windup-rulesets/blob/master/rules/rules-reviewed/azure/springboot/spring-boot-to-azure-eureka.windup.xml) and so on

KaiqianYang:
> I am not sure that technology usage rule only scan JARs. The idea behind is "how can I know AWS is used in this app". If it's a JAR (for example, the AWS Java SDK), good. But it could be a specific file (eg. a `context.xml` file in `META-INF` indicates that the app uses Tomcat), or a specific configuration.
> 
> But I will leave @mrizzi and @PhilipCattanach to answer as I might be wrong.

@agoncal Yes, I think if technology-usage rules scan not only jars but also specific files, specific configurations, that is good. And if currently technology-usage rules focus on scanning jars, we can enhance it by adding rules to detect other signals, such as specific configurations.

PhilipCattanach:
@agoncal @KaiqianYang The technology-usage rules are not at all limited to scanning jar files. The rules are happy to scan any source of configuration data that will surface the use of specific technologies.

KaiqianYang:
> @KaiqianYang - Can you update the rules to use one of permitted values for effort https://access.redhat.com/documentation/en-us/migration_toolkit_for_applications/6.1/html/rules_development_guide/reference_material#how_story_points_are_estimated_in_rules
> 
> Other than that the rules look fine. Thanks.

Thank you @PhilipCattanach , updated, I forgot that effort value is a enum🤣

PhilipCattanach:
Thanks @KaiqianYang - reviewed and tested successfully. There is an unrelated issue that is causing our rulesets integration tests to fail. I'll get that issue fixed then I will merge this PR. Thank you.

PR #971, opened 2023-05-25
[WINDUPRULE-1004] - Introduce JWS technology target and reuse appropiate EAP jakarta rules

carlosthe19916:
https://issues.redhat.com/browse/WINDUPRULE-1004

Adding the JWS targetTechnologies to Jakarta Rules


PhilipCattanach:
@carlosthe19916 
Do we need to change the rulesets that already have jakarta-ee as a target?
If an analysis is ran with the jws6 target alias then the transformer below will ensure the rules that have a target of jakarta-ee will get executed.
https://github.com/windup/windup-rulesets/blob/master/rules/rules-reviewed/jakarta-ee9/jws6.windup.technologytransformer.xml

PR #972, opened 2023-05-25
[WINDUPRULE-1005]-JWS: Upgrade direct dependencies that belong to the groupId org.apache.tomcat

carlosthe19916:
https://issues.redhat.com/browse/WINDUPRULE-1005
Tomcat direct dependency upgrade rules


PhilipCattanach:
Hi @carlosthe19916 
Looking at this PR if 10.1.0 is the minimum Tomcat version required shouldn't the rules be checking for the toVersion 10.0.999 ?

PR #979, opened 2023-06-01
Renumbering Connect rules so it's easier to override

agoncal:
To override a rule we need it's id. This PR sets the ids to a fixed value so it's easier to override and less error prone


mrizzi (approved):
@agoncal thanks a lot :+1:

PR #987, opened 2023-06-07
Overriding rules

agoncal:


agoncal:
@mrizzi managed to pass the tests locally. Should be ok to merge ;o)

PR #988, opened 2023-06-09
Refactor the Redis rule

agoncal:
@mrizzi @PhilipCattanach the Redis rule is the only caching rule that is located under `embedded-framework.windup.xml`, all the other caches are under `embedded-cache-libraries.windup.xml`. So I thought I would move it around. 

My only concern is for the `embedded-framework-technology-usage.windup.xml` file. I would remove the Redis identified technology... but on the other hand, no other cache is identified either (as there is no  `embedded-cache-libraries-usage.windup.xml` kind of file).

WDYT ?


agoncal:
I don't know if this could be related, but the test class `rules/rules-reviewed/technology-usage/tests/embedded-framework-technology-usage.windup.test.xml` does point to `<rulePath>../../cloud-readiness/embedded-cache-libraries.windup.xml</rulePath>`

agoncal:
@PhilipCattanach any progress on this? Is there an associated PR ?

PR #990, opened 2023-06-14
Adding information about testing skipping the URL tests

agoncal:


mrizzi (approved):
@agoncal thanks a lot: docs are never enough :+1:

PR #995, opened 2023-06-21
Check if external process is started in java code

showpune:
Update the windup rule to check if to use Process/ProcessBuilder/ProcessHandle to handle external process


PhilipCattanach:
@m-brophy - I have created Jira https://issues.redhat.com/browse/WINDUPRULE-1024
I think we might want to chnage the effort on both rules in the ruleset.

m-brophy:
I've renumbered the tests because by convention we number the tests to match the rule it's testing. I have no idea why the previous test was numbered incorrectly for this one to follow on from.

m-brophy:
lgtm - when ci tests finish running I'll merge

m-brophy (approved):

PR #1007, opened 2023-07-12
Adds WebSocket protocol to localhost

agoncal:
After `http` and `jdbc` this PR adds `ws` and `wss` protocols to localhost


brunoborges:
@agoncal I noticed that file extensions .css, .ts, .js, .html, .htm are not being processed.

That's where I imagine a lot of ws/wss/http/https URLs may be present.

PR #1008, opened 2023-07-12
Adds a rule about CORBA

agoncal:


m-brophy:
Hi @agoncal the tests are failing on this PR because the test class triggers rule `local-storage-00001` in ruleset `local-storage.windup.xml` and  test `local-storage-test-00001` then fails,presumably because it's looking for the rule to fire51 timesand it's getting 52. 

You could maybe refactor your test and data to a subfolder so they don't trigger the other rules.

agoncal:
I've moved the CORBA data to a `corba` subfolder. Unfortunatelly `local-storage` looks into subfolders too. So I got rid of File access in the CORBA sample.

agoncal:
@m-brophy one test is failing but I'm not sure what it is. In the trace there is `Could not find artifact org.jboss.windup.config:windup-config-impl:jar:6.3.0-SNAPSHOT in central`

m-brophy:
I kicked the test off again and it's passed now. Sometimes the build servers struggle a bit in USA work hours and things timeout etc.

agoncal:
Cool ! So it can be merged ;o)

m-brophy:
> I've moved the CORBA data to a `corba` subfolder. Unfortunatelly `local-storage` looks into subfolders too. So I got rid of File access in the CORBA sample.

The subfolder thing only works if all the rulesets in a folder have their test and data in a corresponding subfolder so each test only acts upon it's own data files- sorry, should have been clearer. In this case you would have had to do the same to local-storage too and that probably would have been enough

m-brophy:
I'll just link the PR to a JIRA and then merge it

PR #1009, opened 2023-07-24
[WINDUPRULE-1012] Add rules for the future deprecation of the finalize() methods

jmle:
The `finalize()` usage search has been implemented as a `{*}.finalize()` javaclass condition to avoid creating a `where` clause with many OR'd conditions inside.

See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-1012).


PhilipCattanach:
@jmle The rules worked fine.
Looking at the https://openjdk.org/jeps/421
This change was introduced in OpenJDK18
So I think the _source_ folder should be openjdk17 not openjdk18
If you can rename that folder then we can get it merged.

PhilipCattanach:
All good - thanks @jmle

PR #1010, opened 2023-07-25
[WINDUPRULE-1014] Add rules for several removed APIs

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUPRULE-1014).

Merge [this](https://github.com/windup/windup-rulesets/pull/1009) first.


PhilipCattanach:
@jmle - Can you split the first rule into two rules as 
java.lang.ThreadGroup.allowThreadSuspension({*})"
Has nothing to do with finalize which is what the issue title and message is referring to.
Also is 1 SP enough, we have allocated 3 SP in the finalize-deprecation ruleset which was the first OpenJDK21 ruleset implemented

PhilipCattanach:
@jmle the final rule is about the removal of java.lang.Compiler and  javax.management.remote.rmi.RMIIIOPServerImpl classes
But the issue title and description is referring to finalization. Please split that rule into 2 rules, one per class and explicitly call of the class names as you have with java.lang.ThreadGroup.allowThreadSuspension

PR #1016, opened 2023-08-03
WINDUPRULE-1018: new ruleset for eap8 log4j removed

m-brophy:


PhilipCattanach:
@m-brophy Thanks for the PR
The message for the first rule has a spelling mistake, and it is also reflected in the tests
Log4j usage without co**m**figuration files
Also the first rule, the logic has back to back **not** statements, shouldn't they be joined by an **and**
It seems to work but the logic is no config files AND no log4j embedded artifact AND references to Log4J

PhilipCattanach:
@m-brophy - The rules look fine.
Only 1 thing that needs to change is the Ruleset description, It currently reads:

This ruleset provides analysis of Java EE applications that need to change their configuration to migrate from Keycloak to native EAP OpenID Connect support 

If you can update that then I'll merge to master

PhilipCattanach:
Tested successfully locally.

PR #1018, opened 2023-08-11
[WINDUP-3420] Remove YAML matching from rules

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUP-3973).

This will prevent these rules from analysing yaml files. At the moment there is no possibility of doing so. A potential solution will have to be investigated.


PhilipCattanach:
The changes look fine and tested successfully

PR #1030, opened 2023-10-17
[WINDUPRULE-1037] Improves EAP 8 JSP rule, clarifying it is about Fac…

emmartins:
…es removing support for JSP, and make it potential
Issue: https://issues.redhat.com/browse/WINDUPRULE-1037

### DIFF
--- a/rules/rules-reviewed/eap8/eap6/eap8-faces.windup.xml
+++ b/rules/rules-reviewed/eap8/eap6/eap8-faces.windup.xml
@@ -178,14 +178,24 @@
         </rule>
         <rule id="eap8-faces-00009">
             <when>
-                <or>
-                    <filecontent filename="{*}.jsp" pattern="&lt;jsp"/>
-                    <filecontent filename="{*}.jsp" pattern="&lt;%"/>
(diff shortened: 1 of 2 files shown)


PhilipCattanach:
@emmartins - You need to update the wording in the test, as you've changed the message.
The test is failing with

Error with test: eap8-faces-00009-test
	Cause: (Rule: eap8-faces-00009-test) Assertion failed: [eap8-faces-00009] The hint was not found!

emmartins:
@PhilipCattanach think I need some help here, I changed the test message but it is still failing, maybe something not good with the rule itself? The idea is to catch  whenever one of the 2 dependencies is present AND there is a jsp file...

emmartins:
all seems good to merge, thanks for the help

PhilipCattanach:
Successfully tested, happy to merge. thanks @emmartins

PhilipCattanach:
There is an integration test that is failing beacsue of a broken link. Nothing to do with this change, so we can address that separately under https://issues.redhat.com/browse/WINDUPRULE-1039

PR #1043, opened 2023-12-05
[WINDUPRULE-1042] Implement JMS to Quarkus rules

jmle:
Extracted from https://github.com/savitharaghunathan/helloworld-mdb/compare/main...quarkus


PhilipCattanach:
@jmle Thank you for the PR.
Overall a great addition but I do have a few comments. 

1). There are duplicate rule ids -00030

2). Some of the rules only contain a link to the https://quarkus.io/guides/ page. 
Can we not be more specific than that?

3). Why do the rules that reference javax.ejb.ActivitionConfigProperty, javax.jms.Queue and javax.jms.Topic not also 
check for the jakarta. equivalents?

PhilipCattanach:
Thanks @jmle - I'll review this PR later today.

PhilipCattanach:
A couple of observations, but nothing to prevent the merge
1). I think there is no need for the apostrophe in the hint title for rule #30 and #40
2). I would have like to have data to trigger the jakarta based when criteria. All of the rules tests are driven by the javax test data.

PR #1044, opened 2023-12-06
[WINDUPRULE-1019] adds eap 8 ruleset deprecated-initialcontextfactory…

emmartins:
…-is-removed
Issue: https://issues.redhat.com/browse/WINDUPRULE-1019


emmartins:
not ready to merge

emmartins:
@PhilipCattanach yes but it should be 3, 2 in the class and 1 in the wsdl, so need something more in the rule, to catch the value reference to the class name.

PhilipCattanach:
@emmartins - Thank you. The rules look fine and my tests were successful. I'll merger to master.

PR #1060, opened 2024-01-19
Package OpenRewrite recipes as jar

ErhardSiegl:
This is a contribution to https://issues.redhat.com/browse/WINDUPRULE-1046 
It enables to use the following plugin configuration in `pom.xml`
```
            <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>5.17.1</version>
                <dependencies>
                    <dependency>
                        <groupId>org.jboss.windup.rules</groupId>
                        <artifactId>windup-rulesets</artifactId>
                        <classifier>rewrite</classifier>
                        <version>6.4.0-SNAPSHOT</version>
                    </dependency>
                </dependencies>
            </plugin>
```
and then e.g:
```
mvn rewrite:discover | grep windup
[INFO]     org.jboss.windup.eap8.FacesWebXml
[INFO]     org.jboss.windup.jakarta.javax.BootstrappingFiles
[INFO]     org.jboss.windup.jakarta.javax.PersistenceXml
[INFO]     org.jboss.windup.JavaxActivationToJakartaActivation
[INFO]     org.jboss.windup.JavaxAnnotationToJakartaAnnotation
[INFO]     org.jboss.windup.JavaxBatchToJakartaBatch
[INFO]     org.jboss.windup.JavaxDecoratorToJakartaDecorator
[INFO]     org.jboss.windup.JavaxEjbToJakartaEjb
[INFO]     org.jboss.windup.JavaxElToJakartaEl
```
The code is taken from https://github.com/quarkusio/quarkus-updates/blob/main/recipes/pom.xml and https://github.com/quarkusio/quarkus-updates/blob/main/recipes/src/main/assembly/core.xml

This is meant as proof of concept and does not fix WINDUPRULE-1046
One problem is, that `windup-cli --openrewrite` uses `-Drewrite.recipeArtifactCoordinates` but this option apparently doesn't understand maven classifiers.
Also it might not be the best solution to generate a separate artifact instead of just moving the `rewrite.yml` files to `META-INF/rewrite`. However that would have more impact, while this change is compatible with older versions.

### DIFF
--- a/rules/pom.xml
+++ b/rules/pom.xml
@@ -384,6 +384,25 @@
                   </includes>
                </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>generate-openrewrite-jar</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                        <configuration>
+                            <descriptors>
+                                <descriptor>src/main/assembly/rewrite.xml</descriptor>
+                            </descriptors>
+                            <appendAssemblyId>true</appendAssemblyId>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
             <!--
             we need to download the CSS files each time we generate the surefire report due to
             https://issues.apache.org/jira/browse/SUREFIRE-616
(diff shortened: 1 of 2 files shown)


m-brophy:
Thanks for this @ErhardSiegl - this is a really good idea and I'll have a think about whether it's best to create the jar or just move the recipes to META-INF/rewrite. Does openrewrite automatically parse *any* yaml files there? Or do they have to be named `rewrite.yml`? I'm just thinking of the logistics of moving several files with the same name  to the same folder.

ErhardSiegl:
The files don't have to be named `rewrite.yml`. Many projects use different names (see e.g. Quarkus) and have it directly under `META-INF/rewrite`.
Indeed, in my example the directory structure is maintained to avoid conflicts.

